PR #68, opened 2020-07-29
feat: move to actix

izik1:
Note: this needs a good review, it touches over a thousand lines of code, and there is probably at least _one_ problem with it


backend/api/crates/server/src/auth.rs
```diff
@@ -100,41 +142,27 @@ pub fn reply_signin_auth(
...
-            warp::reply::with_header(
-                reply,
-                "Set-Cookie",
-                &format!(
-                    "{}={}; Secure; HttpOnly; SameSite=Lax; Max-Age={}; domain={}",
-                    JWT_COOKIE_NAME, jwt, MAX_SIGNIN_COOKIE, COOKIE_DOMAIN
-                ),
-            )
-        }
+    let set_cookie = if SETTINGS.get().unwrap().local_insecure {
```
mehcode, line 147: Use `.add_cookie` with the `Cookie` type instead of `format!`

backend/api/crates/server/src/endpoints/user.rs
```diff
@@ -1,118 +1,140 @@
...
+
+    get_by_id(db.as_ref(), &claims.0.id)
+        .await
+        .map_err(|_| HttpResponse::InternalServerError())?
+        .map(Json)
+        .ok_or(HttpResponse::NotFound().json(NoSuchUserError {}).into())
+}
+
+// fixme: this should be just `GET /user/sso-jwt`
+#[get("/user/get-sso-jwt")]
```
mehcode, line 113: Route ideas. This should not be GET because its a creation.

```
POST /authorize
POST /token
```

PR #71, opened 2020-07-31
make routes work with the new schema

### LINKED ISSUE #40: trait to get path from ApiEndpoint
feature gate it for backend
should have default impl that uses the uri string

See https://github.com/seanmonstar/warp/issues/621

izik1:
Also fixes #40


mehcode:
@izik1 pretty sure this was also in #72 – rebase and re-open if not

PR #101, opened 2020-08-10
feat(backend): remove API-js

izik1:


backend/api/src/jwkkeys.rs
```diff
@@ -0,0 +1,184 @@
...
+            .ok_or_else(|| anyhow!("Missing Key ID"))?;
+
+        let jwk_key = self
+            .keys
+            .get(&token_kid)
+            .ok_or_else(|| anyhow!("Invalid Key ID"))?;
+
+        let claims: Claims = self.decode_token_with_key(jwk_key, token)?.claims;
+        let now = get_current_timestamp();
+        if claims.auth_time > now || claims.iat > now {
```
dakom, line 135: Just curious - is this enough of a claims check? I'm not totally familiar with the JWK workflow, so might be missing something - just kinda comparing to the Node implementation at https://github.com/firebase/firebase-admin-node/blob/24ac03fc9d36ee675719bf64cb1ff8cd7007f39b/src/auth/token-verifier.ts#L196 which seems to step through more of the claims (`iss`, `aud`, etc.)

dakom: @izik1 explained via Slack, that this is dealt with in https://github.com/ji-devs/ji-cloud/pull/101/files#diff-c633486e93699bbc00467a0393ca9d2fR154 (and via the jsonwebtoken crate at https://docs.rs/jsonwebtoken/7.1.1/jsonwebtoken/struct.Validation.html

PR #164, opened 2020-09-22
fix: use the right response type for register

izik1:


dakom:
@mehcode I'm merging this in since it's such a tiny quick change and helps me out on the frontend side. Hope that's okay!

PR #206, opened 2020-10-14
fix: allow for the algolia image index to be chosen manually if running locally

izik1:
Like #202, this requires re-syncing algolia, unlike #202, it needs to do a total re-sync
* the `image` index *should* be removed
the following sql needs to be ran:
```sql
update image_metadata set last_synced_at = null
```


izik1:
Note that this and #202 should preferably be rolled into 1 PR so that we don't have to re-sync things twice

PR #230, opened 2020-10-22
fix: make api compile (sqlx prepare)

izik1:


izik1:
self merging: it's a trivial change and CI doesn't run on PRs

PR #388, opened 2020-11-20
feat: allow for ordering of age_range, affiliation, subject, and style

Sheng-Long:


izik1:
Need to run `cargo sqlx --prepare -- --lib` (the `-- --lib` part is to work around a bug that you'll see if you don't use it), along with some unused imports that should probably be removed

izik1 (approved):

PR #551, opened 2020-12-17
Adding several components

NaomiLea:


dakom (changes requested):
This PR currently checks in almost 2,000 files, due to introducing a new `node_modules 13.04.22` folder. 

We could just .gitignore `node_modules*` but I haven't seen this versioning approach before and it seems like a mistake

PR #574, opened 2020-12-24
committed image delete transaction

romansenin:


izik1:
Thanks for the catch!

PR #584, opened 2020-12-27
Adding component for first image label page

NaomiLea:
Component for first image label page and elements that will be used within that page and other pages.


dakom:
Merging for now. Later, will add testing which will include stricter Typescript checks.

Issue #614, opened 2021-01-07
Login Full One (Storybook) - Reformat line for "Don’t have an account yet? Sign up"


esteels:
Previous edits/issues to Login screens resulted in reformatting of line for "Don’t have an account yet...". Neither the Zeplin old design pages nor the current Storybook screen reflect these changes, but the previous Ji Cloud screens do (see screenshots below). 

Please reformat the line in the following ways:
(1) Add Sign up -- "Don’t have an account yet? Sign up"
(2) Adjust coloring so that "Don't have an account yet?" is regular text, and "Sign up" is blue/hyperlinked text

---

**Previous Ji Cloud Screen reflecting this reformatted text:**

![Registration_28_C_LoginFullOne_AddSignUpLink_Ji](https://user-images.githubusercontent.com/74170693/103942942-35893880-50ff-11eb-8571-8cb361d2d514.png)

---

**Storybook link, without updated text formatting:**

![Registration_28_A_LoginFullOne_AddSignUpLink_SB](https://user-images.githubusercontent.com/74170693/103942939-34f0a200-50ff-11eb-8438-eaae20c81119.png)

---

**Storybook link, without updated text formatting:**

![Registration_28_B_LoginFullOne_AddSignUpLink_Zeplin](https://user-images.githubusercontent.com/74170693/103942941-34f0a200-50ff-11eb-8daf-50e527cb2f29.png)

NaomiLea:
This has been fixed. Just needs to be merged, please check on Monday

esteels:
@NaomiLea, update is not displaying. Please revisit to ensure the following: 

(1) Add Sign up -- "Don’t have an account yet? Sign up"
(2) Adjust coloring so that "Don't have an account yet?" is regular text, and "Sign up" is blue/hyperlinked text

---

**Storybook link, without updated text formatting on 01/12/21:**

![Registration_28_E_LoginFullOne_AddSignUpLink_SB_Round2](https://user-images.githubusercontent.com/74170693/104392903-20524680-5511-11eb-918f-1e4258a1e8f3.png)

esteels:
@NaomiLea, looks good. Thank you!

dani-ji:
@NaomiLea @esteels as the current Zeplin file https://app.zeplin.io/project/5fd249c98c6a325399642201/screen/5fec86fa49183e4e3b3412c4 shows the text differently please follow the Zeplin

esteels:
[...]

PR #637, opened 2021-01-13
Fix formatting and #614

NaomiLea:


NaomiLea:
#614  can be checked @esteels

Issue #681, opened 2021-01-18
QA Publish (Publish Full One) - Add default text [placeholder?] for JIGs Name


esteels:
[Zeplin sketches](https://app.zeplin.io/project/5fd249c98c6a325399642201/screen/5fe1a1570c56693dee495b9c) include default/grey text for each of the fields of Publish Full One screen. Current [Storybook screen](https://storybook.sandbox.jicloud.org/?path=/story/full-pages-publish--publish-full-one) is missing default text (even placeholder) in "JIG's name" field. 

Please add default/grey text for the "JIG's name" field.

---

**Zeplin Sketch (with JIG's name field filled out):**

![Publish_04_A_PublishFullOne_EmptyFieldsText_Zeplin](https://user-images.githubusercontent.com/74170693/104874192-4ec08f00-5920-11eb-93bf-08cdcabdd9d5.png)

---

**Storybook Screen (with empty JIG's name field):**

![Publish_04_B_PublishFullOne_EmptyFieldsText_SB](https://user-images.githubusercontent.com/74170693/104874193-4ec08f00-5920-11eb-85be-34e2da384d65.png)

NaomiLea:
This is ready to be checked

esteels:
@NaomiLea not previewing correctly for me on SB [Publish Full One](https://storybook.sandbox.jicloud.org/?path=/story/full-pages-publish--publish-full-one)

---

![Publish_04_C_PublishFullOne_EmptyFieldsText_SB_Round2](https://user-images.githubusercontent.com/74170693/105453686-8eec7e00-5c4e-11eb-9e52-06ca7aa4bfb4.png)

chayaml:
Default text/ placeholder still missing in "JIG's name" field in [Storybook](https://storybook.sandbox.jicloud.org/?path=/story/entry-jig-publish-pages--publish-full-one).

 Please Add default/ grey text for the "JIG's name" field.

![Screen Shot 2021-02-03 at 9 22 41 AM](https://user-images.githubusercontent.com/78186896/106761422-cd951780-6602-11eb-8935-d6d717c51ef6.png)

NaomiLea:
Fixed

dakom:
duplicate of https://github.com/ji-devs/ji-cloud/issues/1319

PR #719, opened 2021-01-20
#681 Fix and adding publish page back in

NaomiLea:


frontend/elements/src/core/titles/variants/plain-white.ts
```diff
@@ -24,13 +28,15 @@ export class _ extends LitElement {
...
+  small:boolean = false; 
 
   render() {
 
-    const {title, bold} = this;
+    const {title, bold,small} = this;
 
     return html`
-    <p class="${bold ? 'bold' : ''}">${title}</p>
+    <p class="${bold ? 'bold' : ''} ${small ? 'small' : ''}">${title}</p>
```
dakom, line 39: This file shouldn't be used anywhere, instead we have `title-ji` with color options. Please delete these temporary variant references when you've moved it over. If the differences can't be expressed in `title-ji` then we may need new elements, but color alone isn't a problem

frontend/elements/src/core/lists/searchbox-li-check.ts
```diff
@@ -0,0 +1,46 @@
+import { LitElement, html, css, customElement, property } from 'lit-element';
+import {nothing} from "lit-html";
+
+@customElement('searchbox-li-check')
```
dakom, line 4: This name doesn't really make sense. The li element itself is not a "searchbox" the way the li-check element _is_ an element with a checkbox. Please refactor this.

The first option would be to see if you can compose that out of existing elements, with slight changes. If the changes are too proprietary and make usage _elsewhere_ problematic - then it should be split out.

In that case, it makes more sense to move all the "searchbox" stuff into its own widget (e.g. you'd have `searchbox-input`, `searchbox-li`, all contained under that folder - and then an example `searchbox` component)

alternatively, if the design of this element is something which is reusable in contexts other than "searchbox", please rename and express what it _is_ as opposed to where it's _used_

for example - `button-rectangle` is _not_ called `submit-button`

frontend/storybook/src/components/entry/homepage/homepage-search.ts
```diff
@@ -0,0 +1,42 @@
...
+const STR_RED = "red";
+const STR_SEARCH = "Search";
+
+export const Search = (props?: SearchArgs) => {
+
+    const { checked } = props || DEFAULT_ARGS;
+
+    return `
+    <search-bar>
+    <search-dropdown slot="dropdown" placeholder="All ages">
```
dakom, line 29: text here needs to be moved out into a `STR_`

frontend/storybook/src/components/entry/homepage/homepage-search.ts
```diff
@@ -0,0 +1,42 @@
...
+        
+        <searchbox-li-check ${checked && 'checked'}>Passover</searchbox-li-check>
+    
+    </search-dropdown>
+    <search-dropdown slot="dropdown-language" placeholder="All languages"></search-dropdown>
+    <button-rect slot="button" iconBefore="magnifyer" size="medium" bold=true color="${STR_RED}" largetext=true>Search</button-rect>
+    <title-ji slot="advanced" color="lightblue" weight="x-bold">Advanced <br> Search</title-ji>
```
dakom, line 36: If this is meant to be interactive, it should either use `button-text` or `link` attribute on `title-ji`.

If it is not meant to be interactive, it should be in the element instead of the component

frontend/storybook/src/components/entry/homepage/homepage-top.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Search } from "~/components/entry/homepage/homepage-search";
+export default {
+  title: 'Homepage',
+}
+
+export const TopSection = () => {
+    return `
+    <header-blue>
+        <title-ji slot="subtitle" size="x-large" color="lightblue" weight="x-bold">Learning Through</title-ji>
+        <title-ji slot="subtitle" size="x-large" color="lightgreen" weight="x-bold"> Creation</title-ji>
```
dakom, line 12: Static content should be in the element, not the component.

In this case there should be a `homepage-header` element

frontend/storybook/src/components/entry/publish/publish-one.ts
```diff
@@ -0,0 +1,99 @@
...
+const STR_SELECTLIST = "Select from the list";
+const STR_SELECTCATEGORIES ="Select from the categories";
+
+export const PublishFullOne = (props?:PublishArgs) => {
+
+ const {uploaded, errormessage, instruction, errorwrapper, errorname} = props || DEFAULT_ARGS;
+
+
+    return `
+    <publish-page title="${STR_TITLE}" subtitle="${STR_SUBTITLE}">
```
dakom, line 69: static content should be in the element not the component. Applies here and to other titles and spacers below

frontend/storybook/src/components/entry/publish/publish-two.ts
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+export const PublishFullTwo = (props?:PublishArgs) => {
+
+ const {} = props || DEFAULT_ARGS;
+ 
+
+    return `
+    <publish-page-two title="${STR_TITLE}" subtitle="${STR_SUBTITLE}">
```
dakom, line 52: static content should be in the element not the component. applies here and other text elements below

dakom (commented):
@NaomiLea please fix the issues addressed in this review. Thanks

frontend/elements/src/core/inputs/dropdowns/search-dropdown.ts
```diff
@@ -0,0 +1,99 @@
+import { LitElement, html, css, customElement, property } from 'lit-element';
+import { classMap } from 'lit-html/directives/class-map';
+import {nothing} from "lit-html";
+@customElement('search-dropdown')
```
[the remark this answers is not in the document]
NaomiLea: It has a different layout than normal dropdown and it will be a lot more code to actually alter the styling than to have a separate element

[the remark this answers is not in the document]
NaomiLea: this is only connected to homepage so I will move it there.

PR #731, opened 2021-01-20
QA FIXES

NaomiLea:


frontend/storybook/src/components/entry/user/login/pages/landing.ts
```diff
@@ -12,6 +12,7 @@ const STR_PASSWORD = "Password";
 const STR_FORGOTTEN ="Forgot your Password?";
 const STR_USERLABEL = "User Name";
 const STR_SUBMIT = "Submit";
+const STR_PASSWORDPLACEHOLDER = "Type your password";
```
dakom, line 15: _very_ tiny nit - should be `STR_PASSWORD_PLACEHOLDER` with an underscore in between like in `start.ts`.

Not a big deal at all, just noticed since this is one of the things which gets duplicated on the Rust side and it should match to avoid the discrepancy and make "find" more reliable.

PR #762, opened 2021-01-27
Publish elements 

NaomiLea:
Adding publish elements and fixes from QA

### DIFF
--- a/frontend/elements/src/entry/jig/publish/dropdown-list-object.ts
+++ b/frontend/elements/src/entry/jig/publish/dropdown-list-object.ts
@@ -1,6 +1,8 @@
 import { MEDIA_UI } from '@utils/path';
 import { LitElement, html, css, customElement, property } from 'lit-element';
+import { nothing } from "lit-html";
 
+export type Mode = "share" | "embed" | "link";
 @customElement('dropdown-list-object')
 export class _ extends LitElement {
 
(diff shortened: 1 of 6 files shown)


frontend/elements/src/entry/jig/publish/dropdown-list-object.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  render() {
+    const {icon, label} = this;
+   
+   
+
+    return html`
+<div class="wrapper">
+    <div class="inside-wrapper">
+        <img-ui path="${icon}"></img-ui>
```
dakom, line 51: generally speaking, UI images are predefined and should be a variant or hardcoded - e.g. it doesn't make sense to accept any image in our dropbox as a possibility here.

dakom (changes requested):

frontend/elements/src/entry/jig/publish/sharing-caring.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  render() {
+
+    const {sharing} = this;
+    
+    return html`
+    <main class="${sharing ? 'sharing' : ''}">
+    <div>
+       <img-ui class="close" path="icn-x-close.svg"></img-ui>
+       <h1>Sharing is Caring</h1>
```
dakom, line 55: Display strings should be outside the html template

dakom (changes requested):

frontend/elements/src/entry/jig/publish/dropdown-list-object.ts
NaomiLea: If you look at publish-two component you can see how it is being used. Would that be an element in itself?

dakom: The file as it is here is under `publish` so I assume this `dropdown-list-object` is used _only_ on the publish page. If that's the case, `Icon` should be a variant with those options However it seems these icons are also paired with hardcoded strings, so it's not really about the icon, it's about the overall mode, so `type Mode = "share" | "embed" | "link"`

Then, inside of the render function, you derive the icon and string based on the mode. So you ultimately have something like:

```
`export type Mode = "share" | "embed" | "link"

@customElement('dropdown-list-object')
export class _ extends LitElement {
  @property()
  mode: Mode = "share"
  
  render() {
    const {mode} = this;
    
    const iconPath = mode === "share" ? "share.svg"
      : mode === "embed" ? "embed.svg"
      : "link.svg";
      
     const label = mode === "share" ? SHARE_STR
      : mode === "embed" ? EMBED_STR
      : LINK_STR;
      
      return html`
         <p>${label}</p>
         <img-ui path="${iconPath}"></img-ui>
      `;
        
  }
}
```

Sidenote - "share" seems more appropriate than "student" which is why I called it that here

dakom: If `dropdown-list-object` is more generic and used everywhere then I'd divide it into two separate things:

1. A generic element that only has a slot for the image. Idea is it can be used for _any_ image - not just ui
2. This element which works the same way as described with mode, but doesn't have all the HTML/CSS, rather it merely renders the generic element and supplies the label/icon (with the icon being placed in the slot)

NaomiLea: I will change it into mode

PR #764, opened 2021-01-27
Publish page component

NaomiLea:


frontend/storybook/src/components/entry/publish/pages/publish-two.ts
```diff
@@ -1,86 +1,71 @@
...
-//       const STR_EMBED ="Embed this JIG";
-//       const STR_EMBEDICON ="Icn_Embed.svg";
-
-      
+    const DEFAULT_ARGS:PublishArgs = {
+       closed:false,
+       embedclosed:false,
+      }
 
+      const STR_SHARE = "share the JIG";
```
dakom, line 28: Shouldn't these be in an element?

frontend/storybook/src/components/entry/publish/pages/publish-one.ts
```diff
@@ -34,23 +31,19 @@ export default {
...
       }
 
-const STR_TITLE ="Settings and JIG info.";
-const STR_SUBTITLE = "Last step before publishing";
+
 const STR_BTNLABEL = "Publish JIG";
 const STR_IMGTHUMBNAIL = "red-sea-book.png";
 const STR_SLIDERLABEL = "My JIG is public";
 const STR_NAME = "JIG’s name";
 const STR_LANGUAGE = "Language of instructions";
```
dakom, line 39: shouldn't some of these be in an element (like AGE, LANGUAGE)?

dakom (changes requested):

frontend/storybook/src/components/entry/publish/pages/publish-two.ts
NaomiLea: Its the string for a rectangle button so I believe its best to be kept here

frontend/storybook/src/components/entry/publish/pages/publish-one.ts
NaomiLea: these are used for dropdowns that are elements already

Issue #634, opened 2021-01-12
QA Image labelling


NaomiLea:
https://zpl.io/2ZlAPMl

Please bare in mind the the tree doesnt look 100% right. This will be fixed once implementation is done.

dani-ji:
@NaomiLea where is the page in Storybook?

dakom:
We were having back-and-forth to fix up some showstopping issues outside of the code review mechanism. It's fixed enough now (as in, just within the last half hour) to let the PR through for visual QA - but still needs more work before it can be implemented on the Rust side.

Will merge it now

PR #771, opened 2021-01-28
Adding rest of image label and files affecting it

NaomiLea:


frontend/storybook/src/components/entry/admin/images/meta/pages/confirmation.ts
```diff
@@ -0,0 +1,32 @@
+import "@elements/entry/admin/images/meta/page";
+import "@elements/core/titles/variants/underlined-title";
+import "@elements/entry/admin/images/add/add-image";
+import { Rectangle } from "~/components/core/buttons/rectangle";
+
+export default {
+  title: 'Entry/Admin/Images/Meta/Pages',
+}
+
+const STR_LABEL = "Label Images";
```
dakom, line 10: I think this could be better designed as an element, but not going to block the PR on that

frontend/storybook/src/components/entry/admin/images/meta/pages/three.ts
```diff
@@ -0,0 +1,88 @@
...
+import "@elements/core/titles/variants/plain-black-list"; 
+import "@elements/core/titles/ji";
+import { Rectangle } from "~/components/core/buttons/rectangle";
+
+
+export default {
+  title: 'Entry/Admin/Images/Meta/Pages',
+}
+
+const STR_CATEGORIES = "Categories Summary";
```
dakom, line 14: All the static parts should be in the element. The component should *only* need to set strings for the dynamic parts.

dakom (changes requested):

dakom:
related to #634

PR #773, opened 2021-01-31
publish page fixes

NaomiLea:


frontend/elements/src/entry/jig/publish/page-2.ts
```diff
@@ -1,12 +1,12 @@
-import { MEDIA_UI } from '@utils/path';
+import { mediaUi } from '@utils/path';
 import { LitElement, html, css, customElement, property } from 'lit-element';
+import "@elements/entry/home/TOSORT/placeholder";
```
dakom, line 3: jig shouldn't be importing elements from home, if it's common to multiple entries, should probably be in `core` somewhere

frontend/elements/src/entry/jig/publish/page-2.ts
```diff
@@ -88,11 +86,11 @@ export class _ extends LitElement {
     const STR_SUBTITLE ="What would you like to do next?"
 
     return html`    
-    <main>
+    <main style="background-image:url('${mediaUi("Background@2x.jpg")}')">
         <div class="wrapper">
             
             <div class="inside-wrapper">
-                <slot name="animation"></slot>
+            <placeholder-img class="animation"></placeholder-img>
```
dakom, line 93: "placeholder" is an awkward name for this - hard to tell what it's supposed to do by the name... and convention for our images is generally `img-*`

dakom (changes requested):

PR #786, opened 2021-02-02
feat: add tool for resizing/refreshing media (currently images only)

### LINKED ISSUE #784: Image size change
Canvas size should be `1920x1080` (no matter the input - always scales to this size)

For stickers, if width or height is larger than `1440x810`, then it should scale down to that size (maintaining aspect ratio). If it is smaller, then sticker does not need to resize.

Resizing applies only to static images for now, not animations

izik1:
Currently a draft because:
* poor code quality
* there's a bug that can't show up right now (we only have images on master and sandbox currently), but still could happen if that changed. This tool may be used multiple times if we have some other thing we need to use it for.
* Haven't reviewed it very well (written continuing late into the night)

Will close #784


backend/api/src/image_ops.rs
```diff
@@ -2,25 +2,37 @@ use image::{
...
-use shared::domain::{animation::AnimationKind, image::ImageKind};
+use shared::domain::{animation::AnimationKind, audio::AudioKind, image::ImageKind};
 use shared::media::MediaKind as SharedMediaKind;
 
 /// Kinds of media used with the web media library
 #[repr(i16)]
 #[derive(Copy, Clone, Debug, sqlx::Type)]
 pub enum MediaKind {
-    /// Media is a Png, and an Image
+    /// Media is a Png, and an Sticker Image
```
izik1, line 12: Typo: should be `a Sticker`

PR #789, opened 2021-02-02
Category labeling 

NaomiLea:


dakom (changes requested):
In addition to the comments here, the purpose of "Category one" isn't clear (seems like a mishmosh of category and image meta)

And I don't see a way in Storybook to show/hide the menu

dakom:
seems buggy but merging for now anyway

PR #808, opened 2021-02-11
Fixes to login and signup pages

NaomiLea:


NaomiLea:
Only HTML and CSS fixes

PR #822, opened 2021-02-12
Started to add locale functionality

MendyBerger:
The code is in a pretty bad state but I don't think I broke anything
Only works in Chromium based browsers as it relies on the `<dialog>` element, (will probably be replaced with our own custom dialog element)


dakom (commented):
I realize this is a work in progress, so not gonna block on these, but just a couple small early-on points:

1. use `log::info!`, `log::warn!` etc.instead of the `temp_utils` stuff

2. I assume the `resolve_after` and JS helpers in general are just to help debug for now?

MendyBerger:
`resolve_after` is to simulate db interaction by adding a tiny bit of lag, will remove it once we are connected to the db

PR #840, opened 2021-02-22
Locale custom elements

MendyBerger:


frontend/elements/src/entry/admin/locale/landing.ts
```diff
@@ -0,0 +1,206 @@
...
+        // this.dialog.showModal();
+        // cant get @query to work, dono why
+        this.renderRoot.querySelector("dialog")!.close();
+    }
+
+    public render() {
+        return html`
+            <div class="saving-indicator"></div>
+            <select multiple @change="${this.onBundleSelect}">
+                ${this.bundles.map(([bundleName, selected]) => {
```
dakom, line 161: would be better to slot these in so that you can handle the events on the Rust side, instead of relying on the CustomEvent

frontend/elements/src/entry/admin/locale/landing.ts
```diff
@@ -0,0 +1,206 @@
...
+    public dialog!: HTMLDialogElement;
+
+    private onBundleSelect(e: Event) {
+        const select = e.target as HTMLSelectElement;
+        const options = select.options;
+        this.bundles.forEach((b, i) => {
+            b[1] = options[i].selected;
+        });
+        
+        this.dispatchEvent(
```
dakom, line 132: see comment below - basically, it's better to slot the bundles in and avoid the CustomEvent here

dakom:
@MendyBerger - I only took a look at the custom element stuff really. looks good! just see the notes about custom events and slotting in functionality instead

PR #862, opened 2021-03-10
Color select component

MendyBerger:


frontend/apps/crates/components/src/color_select/actions.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub fn add_user_color(state: Rc<State>, color: Color) {
+    let mut user_colors = state.user_colors.lock_mut();
+    let first_empty = user_colors.iter().find_position(|c| c.is_none());
+
+    match first_empty {
+        Some((empty_slot, _)) => {
+            user_colors.set_cloned(empty_slot, Rc::new(Some(UserColor {
+                color,
+                time_created: Utc::now(),
```
dakom, line 15: this is kinda odd... I assume it's a placeholder for now until we have backend capability to stash the colors?

not gonna block the PR on it, just curious

MendyBerger: @dakom  Not sure what part is odd about it, I assume that this component will get passed a `MutableVec` of user colors, if the passer wants to listen for changes to be able to save to the backend they could easily do it by listening to the `MutableVec` signal.

Does this make sense to you or do you want it to be setup differently?

dakom: what is `time_created` used for?

MendyBerger: https://github.com/ji-devs/ji-cloud-flowcharts/blob/master/modules/components/select-color.drawio.svg

image

dakom: The note in [Zeplin](https://zpl.io/VDXNZve) is a bit more clear about this:

> Place holders for added colors.
After the user used the picker to add a color it is added at the end of the ready made colors.
the user has 8 place for additional colors, if he add the 9th color, it replaced the older color added (the first)

so it's not really an issue of comparing timestamps, it's an issue of the current index vs. max index.

In other words, the algorithm is something like this:

1. we have `N_SLOTS` of "user colors" and `INDEX` of where to place the next user color
2. `INDEX` starts at 0
3. Every time a user color is added, `INDEX` is increased by 1
4. If `INDEX == N_SLOTS`, then it's reset to 0

dakom: So if we have 9 slots and we add 12 colors, the value of INDEX will be:

- 0
- 1
- 2
- 3
- 4
- 5
- 6
- 7
- 8
- 0
- 1
- 2

MendyBerger: Good point, didn't think of it in this way before, I guess I'll change it

PR #867, opened 2021-03-12
steps-nav and menu-tabs

MendyBerger:
Button circle was also effected by this commit


frontend/elements/src/core/menu/menu-tabs/menu-tab-title.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export type TitleKind = ""
+    | 'background-image'
+    | 'color'
+    | 'overlay'
+    | 'text'
+    | 'image'
+    | 'audio';
+
+const STR_LABEL_LOOKUP: any = {
```
dakom, line 13: would be slightly nicer to constrain the keys to TitleKind - but not a big deal :)

frontend/elements/src/widgets/nav/step-nav.ts
```diff
@@ -0,0 +1,62 @@
...
+    @property({type: Boolean})
+    active: boolean = false;
+
+    render() {
+        return html`
+            <button-circle
+                label="${this.label}"
+                color="${ifDefined(
+                    this.active ?  "blue"
+                        : this.completed ? "green"
```
dakom, line 53: wasn't aware of ifDefined directive, cool stuff :)

Not for now - but in the future maybe we should have a utility function that wraps around it, maybe with an interface like [Ramda's cond](https://ramdajs.com/docs/#cond)

PR #874, opened 2021-03-16
Poster cover elements

MendyBerger:
This is not 100% complete, I'll do some touchups/additions tomorrow, but I think that its merge ready


dakom:
looks very clean :)

PR #954, opened 2021-04-16
fixup rollup build, allowing external packages

MendyBerger:


MendyBerger:
@dakom  The Jest test has failed but I'm going to merge it anyway since it looks to me like the test is just a placeholder and is not actually testing anything 

The test is: `expect(1+2).toBe(3)`

PR #955, opened 2021-04-19
feat(backend/api): add 204 response for get root URL

rrcwang:


backend/api/src/http.rs
```diff
@@ -204,6 +204,9 @@ pub fn build(
             .configure(endpoints::media::configure)
             .configure(endpoints::session::configure)
             .configure(endpoints::locale::configure)
+            .configure(
+                |cfg| -> () { cfg.route("/", shared::api::Method::Get.route().to(no_content_response)); }
+            )
```
izik1, line 209: prefer `service(<fn>)` and `#[actix_web::get("/")]` in this case- see [`get_spec`](https://github.com/ji-devs/ji-cloud/blob/05deb38f80c8d7d5aa8bb2703c01e2c170f0536a/backend/api/src/http.rs#L227-L232). Alternatively, use the following:
```rust
.route("/", paperclip::actix::web::get().to(<fn>))
```
Note, however, in this case, this can be succulently written with the following 1 liner:
```suggestion
            .route("/", paperclip::actix::web::get().to(|| NoContent))
```

backend/api/src/http.rs
```diff
@@ -238,6 +241,11 @@ fn default_route() -> HttpResponse {
     ))
 }
 
+fn no_content_response() -> HttpResponse {
+    actix_web::HttpResponse::build(http::StatusCode::NO_CONTENT)
+        .finish()
+}
+
```
izik1, line 248: Routes should avoid returning `HttpResponse` (paperclip can't tell what it is), instead, routes that return nothing should use [`paperclip::actix::NoContent`](https://paperclip.waffles.space/paperclip/actix/struct.NoContent.html)
```suggestion
fn no_content_response() -> NoContent {
    NoContent
}
```
Additionally, for consistency, prefer names like `get_base` or `index` over `no_content_response`.

izik1 (changes requested):
Honestly, for a initial pass this isn't bad, formatting failed in CI, and there are some missed idioms, but you'll pick the idioms up over time, and fixing formatting is as easy as `cargo fmt`

rrcwang:
```Rust
.route("/", paperclip::actix::web::get().to(|| NoContent))
```
This didn't compile because using the bare closure didn't satisfy the async `Factory<_,_,_>`. Seems like paperclip doesn't like taking closures because wrapping `NoContent` in an async block wasn't enough to satisfy `Apiv2Operation`:
```Rust
.route("/", paperclip::actix::web::get().to(|| async { NoContent }))
```
I guess this actix idiom using closures (actix/examples#359) doesn't work with paperclip?

izik1:
> I guess this actix idiom using closures ([actix/examples#359](https://github.com/actix/examples/issues/359)) doesn't work with paperclip?

Seems so, sorry about that.

izik1 (approved):

PR #983, opened 2021-04-26
Fix formatting and tests

izik1:
@dakom, @dani-ji I noticed that formatting was failing on sandbox, so I went and fixed that... I also noticed that the struct for memory modules changed, which broke a test on the backend...


dakom:
hmm sorry about that!

PR #1342, opened 2021-08-16
Refactor publish

dakom:
This changes the jig publish flow from:

1. Create State with placeholders
2. Render it / load it (async race)
3. Inside the render logic, deal with placeholders

To:

1. Load the State
2. Render it

It required changing several lines of code. It was almost entirely about _removing_ the nesting, and it was pretty much a repeatable, smooth process - but I still think it's worth it for @MendyBerger to review before merging, in case I didn't see some original intent somewhere.

Note that I believe we can _also_ remove the `Mutable` wrapper for goals / ages, but I didn't do that here.


dakom:
Whups, there's also a change to use `ModuleThumbnail` in this commit... ignore that ;)

MendyBerger (approved):

PR #1415, opened 2021-08-20
fix(backend/api/image-ops): resize sticker logic, unit tests

### LINKED ISSUE #1389: Image sizing logic for Sticker is wrong
This _maximum_ extents of the different images are defined [here](https://docs.sandbox.jicloud.org/crate/src/shared/domain/image.rs.html#34) but the rules are different for each of them

For Stickers, the logic should be that if either dimension is bigger than the box, it _fits_ the box such that neither dimension exceeds it and the largest dimension matches it.

In other words, a large widescreen image will be downsized such that the width is 1440 and the aspect ratio is maintained, while a large profile image will be downsized such that the height is 810 and the aspect ratio is maintained

Currently - it seems that this is not happening, e.g. this image in the global media library is too large: https://uploads.sandbox.jicloud.org/media/global/f8bbd076-eeeb-11eb-9d31-03c20f0ffb51/resized.png

When this is fixed - all the pre-existing images will need to be batch resized, if we need to open a separate issue for that let me know

rrcwang:
resolves #1389


backend/api/src/image_ops.rs
```diff
@@ -63,7 +63,7 @@ pub fn regenerate_images(
         let new_image = match kind {
             ImageKind::Canvas => original.resize_exact(width, height, FilterType::Nearest),
 
-            ImageKind::Sticker if (width <= original.width() && height <= original.height()) => {
+            ImageKind::Sticker if (width >= original.width() && height >= original.height()) => {
```
rrcwang, line 66: turns out none of the images LARGER than the required size were being processed, while the ones smaller than were :)

dakom: @rrcwang Shouldn't it be `||` ? i.e. if _any_ side is larger than the target?

rrcwang: Good question. This is correct and (round-aboutly) implemented as the logical converse of the requirement in #1389

> For Stickers, the logic should be that if either dimension is bigger than the box, it fits the box such that neither dimension exceeds it and the largest dimension matches it.
>
> In other words, a large widescreen image will be downsized such that the width is 1440 and the aspect ratio is maintained, while a large profile image will be downsized such that the height is 810 and the aspect ratio is maintained

Here we check the condition for "not modifying the image", otherwise the `Sticker` which DOES need to be:
```
if original image fits into required dimensions:
    do not resize
else: // at least one of image dims is larger than the required dimensions
    resize image 
```

dakom: ah, I was (mostly) just reading the lines here out of context - thought "original" meant the defined target size. My mistake!

PR #1571, opened 2021-09-14
Module update

### LINKED ISSUE #1537: timestamps for modules
Modules should have `created_at` and `updated_at` timestamps.

- They should be set to the same value when a module is first created
- Only `updated_at` should be changed when a module is updated
- Cover, which is created internally, must have these set too
- These should be returned with [Module requests](https://docs.sandbox.jicloud.org/crate/shared/domain/jig/module/struct.Module.html)

### LINKED ISSUE #1538: cover_modified for Jig responses
This builds on https://github.com/ji-devs/ji-cloud/issues/1537

Jig responses should include a `cover_modified` field which is a boolean determined by:

1. Grab the Jig's Cover module's `updated_at` and `created_at` timestamps
2. `cover_modified` is true of the cover's `updated_at` is more recent than its `created_at`. otherwise, it's false

I do not believe this should be tracked in the DB, just populated in the Rust struct at runtime.

This is necessary so that we know to show some other placeholder image instead of the cover for jig results where the cover has never been edited (otherwise it's just white)

LeahWhalen:
resolves #1537 
resolves #1538


backend/api/migrations/20210915164531_jig_module_fix_timestamps.sql
```diff
@@ -0,0 +1,2 @@
+alter table jig_module alter column updated_at set default now(),
```
rrcwang, line 1: Date in the filename here needs to be updated so migrations in order

backend/api/Cargo.toml
```diff
@@ -15,6 +15,7 @@ actix-rt = "2.2.0"
 actix-service = "2.0.0"
 actix-web = "4.0.0-beta.8"
 actix-web-httpauth = "0.6.0-beta.2"
+indexmap = "=1.6.2"
```
rrcwang, line 18: Fixed in latest commit on `sandbox`, can be removed

rrcwang (changes requested):
Looks great! merge pending rebase

rrcwang (approved):

PR #1573, opened 2021-09-16
Recent image endpoint

### LINKED ISSUE #1485: Recent images, use a single endpoint to create and bump existing images to the top of the list

LeahWhalen:
completed task of reducing api endpoints for recent image

closes #1485


backend/api/src/db/image/recent.rs
```diff
@@ -7,12 +7,12 @@ use shared::{
 use sqlx::PgPool;
 use uuid::Uuid;
 
-pub async fn create(
+pub async fn update(
```
rrcwang, line 10: These three queries should be able to be cleaned up into a single query using postgres's `on conflict on constraint user_recent_image_user_id_image_id_key do ...` syntax.

See 
* https://www.postgresqltutorial.com/postgresql-upsert/
* https://stackoverflow.com/questions/17267417/how-to-upsert-merge-insert-on-duplicate-update-in-postgresql

backend/api/tests/integration/image/recent.rs
```diff
@@ -128,17 +128,19 @@ async fn update() -> anyhow::Result<()> {
 
     let client = reqwest::Client::new();
 
+    // updating
     let resp = client
```
rrcwang, line 132: Since this endpoint now has a response, we should assert the response snapshot after

shared/rust/src/api/endpoints/image/recent.rs
```diff
@@ -54,13 +38,15 @@ impl ApiEndpoint for Create {
 /// * ['BadRequest'](http::StatusCode::BAD_REQUEST) if the request is malformed.
 pub struct Update;
```
rrcwang, line 39: Should rename this (and relevant functions) to `Put` to reflect the change:
* does an "upsert", rather than an update to an existing resource

shared/rust/src/api/endpoints/image/recent.rs
```diff
@@ -54,13 +38,15 @@ impl ApiEndpoint for Create {
...
 
+// TODO: Move ID into request body
+// TODO: grab req/res from above
 impl ApiEndpoint for Update {
-    type Req = ();
-    type Res = ();
+    type Req = UserRecentImageCreateRequest;
+    type Res = UserRecentImageResponse;
```
rrcwang, line 45: Rename structs to reflect change in usage here

rrcwang (changes requested):

shared/rust/src/api/endpoints/image/recent.rs
LeahWhalen: What would be could for names? How about:
Req = UserRecentUpsertRequest
Res = UserRecentImageResponse

rrcwang (approved):

Issue #1465, opened 2021-08-26
Category - add `user_scope`


dakom:
[Category](https://docs.sandbox.jicloud.org/crate/shared/domain/category/struct.Category.html) should be extended to have a user_scopes which is a `Vec<UserScope>`. By default, new categories (and existing categories in the DB) should have this set to an empty vec

[UpdateCategoryRequest](https://docs.sandbox.jicloud.org/crate/shared/domain/category/struct.UpdateCategoryRequest.html) should support replacing this field too

See https://github.com/ji-devs/ji-cloud/issues/1371#issuecomment-902140753 for discussion

PR #1575, opened 2021-09-17
Rs/user scope

RickyLB:
closes issue #1465


backend/api/src/db/category.rs
```diff
@@ -30,6 +33,15 @@ order by index
         children: vec![],
         image_count: it.image_count as u64,
         jig_count: it.jig_count as u64,
+        user_scopes: {
+            let scopes = it
+                .user_scopes
+                .into_iter()
+                .map(|x| UserScope::try_from(x).unwrap())
```
rrcwang, line 40: `.execpt()` instead of `.unwrap()` would be appropriate for potential debugging purposes

backend/api/src/db/category.rs
```diff
@@ -30,6 +33,15 @@ order by index
...
+                .collect();
+
+            scopes
+        },
```
rrcwang, line 44: style comment: this and lines 77-82 do the same thing. Would prefer using the latter for both as it avoids creating a new scope and declaring a short lived variable

backend/api/src/db/category.rs
```diff
@@ -61,6 +74,12 @@ order by t.ord
         children: vec![],
         image_count: it.image_count as u64,
         jig_count: it.jig_count as u64,
+        user_scopes: {
+            it.user_scopes
+                .iter()
+                .map(|x| UserScope::try_from(*x).unwrap())
```
rrcwang, line 80: style comment: would prefer `.into_iter()` to indicate that ownership is being taken over the entire array, rather than `*` which implies copy of the individual elements

backend/api/migrations/20210915232840_category_user_scope.sql
```diff
@@ -0,0 +1 @@
+alter table category add column user_scopes smallint[] not null default '{}';
```
rrcwang, line 1: 1. Migration should be dated so that it is run last. Other migrations have since been added to the repo since this was created locally
2. Missing a newline char at the end of file

backend/api/fixtures/4_category_nesting.sql
```diff
@@ -1,8 +1,8 @@
...
+insert into category (id, parent_id, name, user_scopes, index, created_at)
+values ('01cff7d8-e910-11ea-8281-7f86c625a156', 'e315d3b2-e90f-11ea-8281-73cd69c14821', 'A.1', (1,4,3), 0,
         '2020-08-28T09:22:34.737562Z'),
-       ('afbce03c-e90f-11ea-8281-cfde02f6b582', null, 'Parent', 0, '2020-08-28T09:20:17.039205Z'),
+       ('afbce03c-e90f-11ea-8281-cfde02f6b582', null, 'Parent', (1,6), 0, '2020-08-28T09:20:17.039205Z'),
        ('e315d3b2-e90f-11ea-8281-73cd69c14821', 'afbce03c-e90f-11ea-8281-cfde02f6b582', 'A', 0,
         '2020-08-28T09:21:43.186002Z'),
-       ('ee749392-e90f-11ea-8281-67d19ffa6107', 'afbce03c-e90f-11ea-8281-cfde02f6b582', 'B', 1,
+       ('ee749392-e90f-11ea-8281-67d19ffa6107', 'afbce03c-e90f-11ea-8281-cfde02f6b582', 'B', (6,7,8), 1,
         '2020-08-28T09:22:02.261811Z');
```
rrcwang, line 8: query is malformed, arrays in postgres should be `array [1, 2, 3]`. `(1, 2, 3)` represents a row/record

rrcwang (changes requested):

Issue #1097, opened 2021-06-08
Likes on jigs


MendyBerger:
User should be able to like  a JIG.
When loading a jig, user should be able to see if they liked this jig already.
When searching Jigs, include total amount of likes on each Jig.

Also need to have the play counts on JIGs.

Both LIKES and PLAYS should be displayed on the card of the JIG search results.

![image](https://user-images.githubusercontent.com/16290361/142207948-5e4f7008-6b84-43f6-bd73-7135ceba9ebe.png)

PR #1930, opened 2021-11-19
feat(be/jig_metrics): add play and like to jigs

RickyLB:
closes issue #1097

### DIFF
--- a/backend/api/migrations/20211118215109_jig_liked.sql
+++ b/backend/api/migrations/20211118215109_jig_liked.sql
@@ -1,0 +1,32 @@
+-- Jig Liked Count 
+alter table jig
+    add column liked_count int8 not null default 0 check ( liked_count >= 0 );
+
+create table jig_like
+(
+    jig_id     uuid              not null          references jig(id) on delete cascade,
+    user_id    uuid              not null,
+    created_at timestamptz       not null          default now(),
+    primary key (jig_id, user_id)
+);
+
+create or replace function update_jig_likes()
+    returns trigger as
+$$
+begin
+    update jig
+    set liked_count = liked_count + 1
+    where id = NEW.jig_id;
+    return NEW;
+end;
+$$
+    language plpgsql;
+
+create trigger add_jig_likes
+    after insert
+    on jig_like
+    for each row
+execute function update_jig_likes();
+
+
+
--- a/backend/api/sqlx-data.json
+++ b/backend/api/sqlx-data.json
@@ -78,6 +78,18 @@
       "nullable": []
     }
   },
+  "04fb89e42da5ea9503934560463dca443ce0dbb9ba2fc38977da4aeb355f8d77": {
+    "query": "\nupdate jig_play_count\nset play_count = play_count + 1\nwhere jig_id = $1;\n            ",
+    "describe": {
+      "columns": [],
+      "parameters": {
+        "Left": [
+          "Uuid"
+        ]
+      },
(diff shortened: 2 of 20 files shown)


johnnynotsolucky:
@RickyLB I'm looking at implementing #1505 and I have a few questions about the API.

The [`<jig>/like`](https://github.com/ji-devs/ji-cloud/pull/1930/files#diff-f9eb12329813b38c23953789b64984c8a5872763e6f8d35f31d527d0181f991aR1085-R1127) endpoint adds a row to `jig_like`, however the dataset which is used to populate `JigResponse` uses the `liked_count`, [example](https://github.com/ji-devs/ji-cloud/pull/1930/files#diff-f9eb12329813b38c23953789b64984c8a5872763e6f8d35f31d527d0181f991aR340-R346) field.

- Should the `liked_count` and `play_count` fields be removed, and instead fetch the count from `jig_like`?
- How should we handle working out if a user has liked a jig:
  - We could include a list of of uids of liked jigs in the response to /user/profile,
  - Or, make an extra request when loading a jig to fetch whether the current user liked the current jig,
  - Or, include a flag indicating whether the current user liked the jig in the JigResponse (this probably not a great idea)

RickyLB:
@johnnynotsolucky 

>  Should the liked_count and play_count fields be removed, and instead fetch the count from jig_like?

Though this does work and yields less code management, this would add more overhead if we had to count every row in `jig_like` especially with a larger user base when querying from db. 

> How should we handle working out if a user has liked a jig? 

Actually, adding a `liked` flag to the `JigResponse` would make most sense if this is just for letting the current user know they have liked the jig in the UI.

johnnynotsolucky:
🙈 maybe I shouldn't have skipped over the migration, sorry!

> Though this does work and yields less code management, this would add more overhead if we had to count every row in jig_like especially with a larger user base when querying from db.

That makes sense.

> Actually, adding a liked flag to the JigResponse would make most sense if this is just for letting the current user know they have liked the jig in the UI.

I think I was worried about adding request-specific data into the response in case it makes caching harder? But this makes sense and is definitely probably the easiest.

Follow up Q: Would it make sense to have play_count work the same as liked_count - instead of using a subquery, have a `play_count` column, and instead of incrementing the row in `jig_play_count`, increment the column in `jig`?

Sorry for all the questions!

MendyBerger:
We already have some request specific data (`author_name`) so I think the ship has sailed on this one

johnnynotsolucky:
Hmmm @MendyBerger wouldn't that be the same regardless of who made the request? It should be from the jig table (or whichever table stores that info).

RickyLB:
@johnnynotsolucky Thanks for being patient for a response!

> Follow up Q: Would it make sense to have play_count work the same as liked_count - instead of using a subquery, have a play_count column, and instead of incrementing the row in jig_play_count, increment the column in jig?

There's a separate table for `jig_likes` so the api can check if a user has already liked a jig or not. From my knowledge `play_count` will increase regardless of who plays it and how many times, so just decided to increment it in a subquery. 

> I think I was worried about adding request-specific data into the response in case it makes caching harder and the live and search endpoints don't require auth, so maybe it would be strange including a field for user data?

I will say I'm still learning the ways of the front end, so I'm not sure how much optimization this would bring. I'm happy to hear what your thoughts are on this so for my learning purposes. I do see what you mean pertaining to caching though. If an api that lists the Uuid's that liked the jig works best, I can do this. What do you think?

Issue #1372, opened 2021-08-18
Search bar on search results pages should be a smaller panel


corinnewo:
At the moment, the search bar remains the same as on the home page when you have search results.
It should actually be a smaller version: see https://zpl.io/VDX9Jwr

corinnewo:
This results page should be the target of the CONTENT menu link.
Search for everything.

MendyBerger:
Partially resolved in 0375f7c8

PR #1956, opened 2021-12-03
fix(fe/home): Active content header for search mode

johnnynotsolucky:
- Possibly related to https://github.com/ji-devs/ji-cloud/issues/1372#issuecomment-905301246
- Implement `From<>` for `PageLinks` so that when navigating to the search page, the Content header is marked as active;
- **Note** I'm assuming that we would want the header to be marked active, but this may not be the case.

image

### DIFF
--- a/frontend/apps/crates/entry/home/src/home/dom/mod.rs
+++ b/frontend/apps/crates/entry/home/src/home/dom/mod.rs
@@ -18,7 +18,9 @@ mod search_section;
 
 pub fn render(state: Rc<State>, auto_search: bool) -> Dom {
     html!("home-full", {
-        .child(page_header::dom::render(Rc::new(page_header::state::State::new()), None, Some(PageLinks::Home)))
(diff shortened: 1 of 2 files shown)


frontend/apps/crates/entry/home/src/home/dom/mod.rs
```diff
@@ -18,7 +18,9 @@ mod search_section;
 
 pub fn render(state: Rc<State>, auto_search: bool) -> Dom {
     html!("home-full", {
-        .child(page_header::dom::render(Rc::new(page_header::state::State::new()), None, Some(PageLinks::Home)))
+        .child_signal(state.mode.signal_ref(|mode| {
+            Some(page_header::dom::render(Rc::new(page_header::state::State::new()), None, Some(PageLinks::from(mode))))
```
johnnynotsolucky, line 22: This could also be 

```suggestion
            Some(page_header::dom::render(Rc::new(page_header::state::State::new()), None, Some(mode.into())))
```

Issue #1856, opened 2021-11-08
Jig Panel - Kebab menu - Can't delete an empty Jig/ module


chayaml:
In the Kebab menu, I am only able to delete a module that has an activity dragged into it. The delete button for an empty module does not delete. 

![Screen Shot 2021-11-08 at 1 23 44 PM](https://user-images.githubusercontent.com/78186896/140797038-1e244416-8005-439b-be8d-4f98a825e447.png)

MendyBerger:
Closing in favor of #1959

PR #1967, opened 2021-12-07
fix(fe/jig/edit): Hide module buttons when they aren't required

### LINKED ISSUE #1959: Last module should not have the menu
There is always one empty module at the end, that module should not have the menu.
The module that's one before that should not have a add (+) button

image

johnnynotsolucky:
Resolves #1959 

This PR makes a few changes to how the kebab menu and add button are rendered in the edit jig sidebar:

- The kebab button is displayed for all modules _except_ for the last one **if** the last one is a placeholder;
- The add button is displayed for all modules _except_ if the module is a placeholder, or the module immediately after is a placeholder;
- The delete button works for placeholders that appear between other modules (#1856).

image

### DIFF
--- a/frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
+++ b/frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
@@ -115,7 +115,7 @@ pub fn delete(state: Rc<State>) {
                 Err(_) => {}
             }
         } else {
-            // The module is DragDrop, it is not persisted so it can be removed from the list with
+            // The module is placeholder, it is not persisted so it can be removed from the list with
             // no extra work required.
             state.sidebar.modules.lock_mut().remove(index);
         }
--- a/frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
+++ b/frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
@@ -114,6 +114,10 @@ pub fn delete(state: Rc<State>) {
                 },
                 Err(_) => {}
             }
+        } else {
+            // The module is DragDrop, it is not persisted so it can be removed from the list with
+            // no extra work required.
+            state.sidebar.modules.lock_mut().remove(index);
         }
     }));
 }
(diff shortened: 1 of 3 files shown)


frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/state.rs
```diff
@@ -45,6 +45,26 @@ impl State {
             Some(module) => module.kind.as_str(),
         }
     }
+
+    pub fn can_add(&self) -> bool {
```
johnnynotsolucky, line 49: I'm not convinced that State is the best place for this logic. Alternatively it could probably go in `module/dom.rs`

frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
```diff
@@ -144,7 +148,17 @@ pub fn assign_kind(state: Rc<State>, kind: ModuleKind) {
...
+
+                {
+                    // Instead of replacing the module at the index, we remove the old module and
+                    // add the new one. This is slightly less efficient because it fires signals
+                    // for the entire list of modules, however, it is necessary so that the modules
+                    // before and after this one can have their views updated.
+                    let mut modules = state.sidebar.modules.lock_mut();
+                    modules.remove(index);
+                    modules.insert_cloned(index, module);
+                }
```
johnnynotsolucky, line 160: this change was necessary to hide the add button for modules when the next module in the list is a placeholder. 

If it's not necessary to hide the add button in those cases, this change can be reverted and probably [this code](https://github.com/ji-devs/ji-cloud/pull/1967/files#diff-78f7288368221bbdbdf08c195a5678f38c88c7f3dfa25723e664c25ccdd2cf2dR49-R66) can be simplified or removed.

frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/state.rs
MendyBerger: That makes sense, you're welcome to do so

frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
MendyBerger: I think that this change is good even if it adds a bit of complexity

frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/dom.rs
```diff
@@ -154,44 +154,56 @@ impl ModuleDom {
...
-                                v.push(MenuDom::item_move_down(menu_state.clone(), state.clone()));
-                                v.push(MenuDom::item_duplicate(menu_state.clone(), sidebar_state.clone(), module.id));
-                            }
-                            v.push(MenuDom::item_delete(menu_state.clone(), state.clone()));
-                            if let Some(module) = &*module {
-                                v.push(MenuDom::item_copy(menu_state.clone(), sidebar_state.clone(), module.id));
-                                v.push(MenuDom::item_duplicate_as(menu_state.clone(), sidebar_state.clone(), module));
+                    // If the module is anything other than a DragDrop, otherwise if it is not the
+                    // last module in the list.
+                    if (&*module).is_some() || (index <= total_len - 2 && (&*module).is_none()) {
```
MendyBerger, line 159: Minor nitpick, I don't think that you need to reference and deference, I think that the compiler handles that automatically when calling a method

MendyBerger (approved):

MendyBerger:
I think this looks great! Thanks @johnnynotsolucky!

PR #1977, opened 2021-12-08
fix(fe/jig/edit): Ensure a new placeholder module is always available in the sidebar

johnnynotsolucky:
Fixes an issue where the placeholder module at the end of the module list in the sidebar wouldn't be available sometimes.

Issue noted in [Slack](https://jewishinteractive.slack.com/archives/C02MDHBQ8EP/p1638870743078500).

To reproduce, add a new module to the list a few times only using the automatically created placeholder module - around the 2nd or 3rd add, it stops appearing.

### DIFF
--- a/frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
+++ b/frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
@@ -22,10 +22,6 @@ pub fn on_module_kind_drop(state: Rc<State>, module_kind: ModuleKind) {
         if state.module.is_none() {
             assign_kind(state.clone(), module_kind);
 
-            // if this is the empty module at the end
-            if state.sidebar.jig.jig_data.modules.len() == state.index {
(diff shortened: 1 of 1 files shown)


frontend/apps/crates/entry/jig/edit/src/edit/sidebar/module/actions.rs
```diff
@@ -157,6 +153,23 @@ pub fn assign_kind(state: Rc<State>, kind: ModuleKind) {
...
                     modules.remove(index);
                     modules.insert_cloned(index, module);
+
+                    // Only add a new placeholder module once the above request has completed and
+                    // the new module has been added to the list of modules.
+                    let placeholder_exists = {
+                        match modules.last() {
+                            // If the list of modules is not empty and the last module is None, then it is
+                            // a placeholder module.
+                            Some(module) => module.is_none(),
```
johnnynotsolucky, line 163: `module` in this case is `Rc<Option<_>>` so we can't use a pattern like `Some(None)`

PR #1989, opened 2021-12-09
chore(fe/jig/edit): Remove sidebar/module/menu/dom which is unused

johnnynotsolucky:
This file is not used anywhere. @MendyBerger I think it's safe to remove?


MendyBerger:
If it's not used than sure

PR #2000, opened 2021-12-13
feat(fe/jig/play): Implement like action on jigs

johnnynotsolucky:
Part of #1505 

- Implements the like endpoint when clicking on the like sidebar action

This works and can be merged. However, for the full feature to be complete, we require some way of knowing whether the current user has already liked a jig.


frontend/apps/crates/entry/jig/play/src/player/sidebar/dom/like.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use super::super::state::State;
+
+pub fn render(state: Rc<State>, jig: &JigResponse) -> Dom {
+    html!("jig-play-sidebar-action", {
+        .property("slot", "actions")
+        .property("kind", "like")
+        // TODO Render active or not active
+        .event(clone!(state, jig => move |_: events::Click| {
+            state.loader.load(async move {
```
MendyBerger, line 23: Something to consider but might not be a problem at all:
loader can only load one future at a time, so if you call `loader.load` when something is already loading the first one will be dropped, so make sure that it's not used anywhere else that could potentially overlap here.

Now that I think about it, this problem could cause some very real issues, we should try coming up with something better (later).

frontend/apps/crates/entry/jig/play/src/player/sidebar/dom/mod.rs
```diff
@@ -45,10 +46,15 @@ pub fn render(state: Rc<State>) -> Dom {
                 state.sidebar_open.set(false);
             }))
         }))
-        .child(html!("jig-play-sidebar-action", {
-            .property("slot", "actions")
-            .property("kind", "like")
-        }))
+        .child_signal(state.player_state.jig.signal_ref(clone!(state => move |jig| {
```
MendyBerger, line 49: Not sure that `published_at` is a reliable way to check if something is published, rather use `draftOrLive` to check.

Checking `published_at` used to be the preferred way to check if something is published but no longer is, so if you took this idea from somewhere else, then please create an issue for that.

johnnynotsolucky: Oh great! I replicated the functionality for rendering the share button: https://github.com/johnnynotsolucky/ji-cloud/blob/sandbox/frontend/apps/crates/entry/jig/play/src/player/sidebar/dom/mod.rs#L52-L60

Will update that too.

frontend/apps/crates/entry/jig/play/src/player/sidebar/dom/like.rs
johnnynotsolucky: @MendyBerger I saw on the profile page when editing details that running requests are cancelled when a new one is made. I had (incorrectly) assumed that it would only cancel the same type of request.

I agree. Perhaps as an interim solution we can remove the [abort call](https://github.com/dakom/dominator-helpers/blob/master/src/futures.rs#L42) from AsyncLoaders replace method?

johnnynotsolucky:
Made this a draft so I can work on it alongside https://github.com/ji-devs/ji-cloud/pull/2016

johnnynotsolucky:
Closing this so that I can work on it with the branch for #2016 as my base.

PR #2016, opened 2021-12-15
feat(be/jig): Implement endpoint for returning whether a jig is liked

johnnynotsolucky:
Part of #1505

- Implements `GET /v1/jig/<id>/like` which returns whether the current user has liked the JIG;
- Updates unlike to `DELETE /v1/jig/<id>/like` instead of `/unlike` so that it indicates the endpoints operate on the same resource.

@RickyLB to update sqlx-data.json I ran:

```bash
$ cargo sqlx prepare -- --lib
$ cargo sqlx prepare -- --test integration
```

I'm not 100% sure why it reshuffled a lot of the queries, maybe it's because my editor removed a bunch of whitespace from `db/jig.rs`?

### DIFF
--- a/backend/api/sqlx-data.json
+++ b/backend/api/sqlx-data.json
@@ -78,6 +78,22 @@
       "nullable": []
     }
   },
+  "0452f5078544e868686ddee838d64a9864b8ab4b6b6546eb3136c4d84662f176": {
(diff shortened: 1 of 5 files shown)


johnnynotsolucky:
WIP for implementing the like feature on the frontend based off this branch: https://github.com/johnnynotsolucky/ji-cloud/pull/1

RickyLB (changes requested):

PR #2032, opened 2021-12-16
fix(fe/register): Send correct location and language data during registration

### LINKED ISSUE #2030: My Profile - Not displaying the location and language of communication I set at registration
I set up my location and language during the registration process, but when I go into the 'My Profile' section, it is not displaying what I had set. Should show what I chose previously.

image

johnnynotsolucky:
Resolves #2030

- Adds a new trait, `Item` which exposes `value()` and `label()` methods;
- Updates `SimpleSelect` so that it can accept a list of a `Item` instead of types which impl `Display`
- Updates the registration process to send the language code which is expected by the profile page instead of the language string.
- Resolves the location not displaying in profile by sending the correct raw location data to the server.

*Note* that this also updates the the available languages in Step 2 to be English, Hebrew and French which is what is available on the profile page. Whereas originally there was English, Hebrew, Spanish, French and Russian. I wasn't sure which was correct so I used what was in use on the profile page.

### DIFF
--- a/frontend/apps/crates/components/src/input/simple_select/mod.rs
+++ b/frontend/apps/crates/components/src/input/simple_select/mod.rs
@@ -3,3 +3,18 @@ mod state;
 
 pub use dom::*;
 pub use state::*;
+
+pub trait Item: Clone {
+    fn value(&self) -> &str;
+    fn label(&self) -> &str;
+}
+
+impl Item for &str {
+    fn value(&self) -> &str {
+        self
+    }
+
+    fn label(&self) -> &str {
+        self
+    }
+}
--- a/frontend/apps/crates/components/src/input/simple_select/mod.rs
+++ b/frontend/apps/crates/components/src/input/simple_select/mod.rs
@@ -4,12 +4,12 @@ mod state;
 pub use dom::*;
 pub use state::*;
 
-pub trait Item: Clone {
+pub trait SimpleSelectItem: Clone {
     fn value(&self) -> &str;
     fn label(&self) -> &str;
 }
 
-impl Item for &str {
+impl SimpleSelectItem for &str {
     fn value(&self) -> &str {
         self
     }
--- a/frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
+++ b/frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
@@ -15,7 +15,7 @@ use utils::{events, languages};
 
 use super::Language;
 
-use components::input::simple_select::{SimpleSelect, Item};
+use components::input::simple_select::{SimpleSelect, SimpleSelectItem};
 
 pub struct Step2Page {}
 
--- a/frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
+++ b/frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
@@ -11,9 +11,11 @@ use crate::{
     },
     strings::register::step_2::*,
 };
-use utils::events;
+use utils::{events, languages};
 
-use components::input::simple_select::SimpleSelect;
+use super::Language;
+
+use components::input::simple_select::{SimpleSelect, Item};
 
 pub struct Step2Page {}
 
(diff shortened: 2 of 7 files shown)


frontend/apps/crates/components/src/input/simple_select/mod.rs
```diff
@@ -3,3 +3,18 @@ mod state;
 
 pub use dom::*;
 pub use state::*;
+
+pub trait Item: Clone {
+    fn value(&self) -> &str;
+    fn label(&self) -> &str;
+}
+
+impl Item for &str {
```
johnnynotsolucky, line 12: This is useful for using lists of &str such as https://github.com/johnnynotsolucky/ji-cloud/blob/sandbox/frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs#L57

frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
```diff
@@ -38,9 +40,9 @@ impl Step2Page {
                         Some(STR_LANGUAGE_LABEL),
                         Some(STR_LANGUAGE_PLACEHOLDER),
                         None,
-                        STR_LANGUAGE_OPTIONS.to_vec(),
+                        languages::EMAIL_LANGUAGES.iter().map(|l| Language::from(l.clone())).collect(),
```
johnnynotsolucky, line 43: Convert into the wrapper for `utils::languages::Language`

frontend/apps/crates/entry/user/src/register/pages/step_3/actions.rs
```diff
@@ -69,11 +69,9 @@ pub fn submit(state: Rc<State>) {
         affiliations,
         location: step_2
             .location_json
-            .map(|raw| serde_json::to_value(JsonRaw { raw }).unwrap_throw()),
+            .map(|raw| serde_json::to_value(raw).unwrap_throw()),
```
johnnynotsolucky, line 72: Resolves the location data.

frontend/apps/crates/components/src/input/simple_select/mod.rs
MendyBerger: :+1:

frontend/apps/crates/entry/user/src/register/pages/step_3/actions.rs
MendyBerger: Just making sure I understand correctly, this change isn't really related to the new trait, right?

frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
MendyBerger: Converting from one type of `Language` to another feels kind of weird but I don't see a way around it, so I guess it's fine

frontend/apps/crates/components/src/input/simple_select/mod.rs
```diff
@@ -3,3 +3,18 @@ mod state;
 
 pub use dom::*;
 pub use state::*;
+
+pub trait Item: Clone {
```
MendyBerger, line 7: I think that the name is a bit too generic, I'd name it something like `SimpleSelectItem` but I don't have super strong feelings about this

MendyBerger (approved):
@johnnynotsolucky I think this is great!
If you wanna make any changes than go ahead, otherwise you can just merge

frontend/apps/crates/entry/user/src/register/pages/step_3/actions.rs
johnnynotsolucky: Yeah, this resolves the location not being shown in the users profile.

frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
johnnynotsolucky: I'm not super happy about using it either. I tried going with having the lists convert to a hashmap and SimpleSelect using Entry but it made the SimpleSelect quite a bit more complex. We could look into refactoring a bit at a later stage to improve it.

PR #2072, opened 2021-12-22
feat(fe/components): Navigate user to Image tab when word list is created

### LINKED ISSUE #1847: Create: Word List - switching between text tab and Image tab before clicking "done"
Currently, when a user creates a word list in the **text** tab, if they press on the **image** tab (without clicking "done"), and then go back to the **text** tab, they lose the words that they just inserted. A user may not realize that they need to hit "done" to save their data and may click on the image tab just to check it out.

Is it possible to save the list so a user can switch off between the tabs without it getting deleted?

Screen Shot 2021-11-07 at 11 38 04 PM

johnnynotsolucky:
Closes #1847

- When a user completes a list and clicks "Done", they are taken straight to the Images tab;
- Image tab is disabled when there is an empty list;
- Cyclic reference exists for widgets so that they can access step_1 state.

### DIFF
--- a/frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_1/state.rs
+++ b/frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_1/state.rs
@@ -18,17 +18,26 @@ use crate::{
 };
 use dominator::clone;
 use futures_signals::signal::Mutable;
+use once_cell::sync::OnceCell;
 use shared::domain::jig::module::body::{Image, _groups::cards::Mode};
 use std::rc::Rc;
 
 pub struct Step1<RawData: RawDataExt, E: ExtraExt> {
     pub base: Rc<CardsBase<RawData, E>>,
-    pub widget: Rc<Widget>,
+    pub widget: OnceCell<Widget>,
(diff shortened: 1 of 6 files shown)


frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_1/state.rs
```diff
@@ -18,38 +18,48 @@ use crate::{
...
 impl<RawData: RawDataExt, E: ExtraExt> Step1<RawData, E> {
     pub fn new(base: Rc<CardsBase<RawData, E>>, tab_index: Mutable<Option<usize>>) -> Rc<Self> {
+        let state = Rc::new(Self {
+            base: base.clone(),
+            widget: OnceCell::default(),
+            tab_index,
+        });
+
+        // Widgets require a reference to the top-level state so that they can have access to any
+        // fields they might require in callbacks.
```
johnnynotsolucky, line 40: This comment relates specifically to [this change](https://github.com/ji-devs/ji-cloud/pull/2072/files#diff-a6de13114d3690edd188660e6ee5f53a9003eb914845d9918f2a17d978dd88a0R132-R134)

PR #2104, opened 2021-12-28
fix(fe/admin/categories): Fix context menu alignment; Re-order items

### LINKED ISSUE #2099: Admin- edit categories - menu needs to be fixed
When you edit a category, the 3-dot menu needs to be fixed:

1. Separate the 'add' and 'delete' items. 
2. I would suggest putting the 'delete' in red as the last item, keeping with how we have delete in other places. 
3. We also usually capitalise the first word in a menu item. Please change to 'Add', 'Move up', 'Move down', 'Rename', 'Delete'

image

johnnynotsolucky:
Part of #2099

@corinnewo this PR 

- Reorders category context menu - delete last;
- Delete item is red;
- Enables editing of any new category - previously you would need to have added a category and then renamed it;
- When renaming a category, the original text is selected so that it is easier to overwrite it

image

image

### DIFF
--- a/frontend/elements/src/core/inputs/primitives/text-content.ts
+++ b/frontend/elements/src/core/inputs/primitives/text-content.ts
@@ -112,22 +112,27 @@ export class _ extends LitElement {
 
     firstUpdated(_changed: any) {
         this.resizeInput();
+        this.handleEditing();
     }
     updated(changed: any) {
         if (typeof changed.get("editing") === "boolean") {
-            const { editing } = this;
-            this.removeGlobalListener();
-            if (editing) {
-                window.addEventListener("mousedown", this.onGlobalMouseDown);
-                const input = this.shadowRoot?.getElementById(
-                    "input"
-                ) as HTMLInputElement;
-                if (input) {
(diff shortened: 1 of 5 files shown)


frontend/elements/src/core/inputs/primitives/text-content.ts
```diff
@@ -112,22 +112,27 @@ export class _ extends LitElement {
 
     firstUpdated(_changed: any) {
         this.resizeInput();
+        this.handleEditing();
```
johnnynotsolucky, line 115: This element can start in the editing state now.

frontend/elements/src/core/menu/ellipses.ts
```diff
@@ -43,6 +43,12 @@ export class _ extends LitElement {
                     box-shadow: 0 3px 16px 0 rgba(0, 0, 0, 0.25);
                     background-color: #ffffff;
                 }
+                ::slotted([slot="menu-content"]) {
```
johnnynotsolucky, line 46: Fixes the alignment of the menu items.

PR #2117, opened 2021-12-30
feat: add icon to empty collapsed module in jig/edit sidebar

mastapegs:
This PR currently uses the wrong image resource.

@MendyBerger, how can I view the images/media available in the media cloud, `https://media.jicloud.org/`?

As of now though, the logic is in place for expanded and collapsed module image rendering, displaying no image for an expanded sidebar, and showing the image for the module when the sidebar is collapsed only.

Once I know the name of the correct resource in our media cloud, I'll replace it with the correct path and this should be good to go!

Collapsed:
image

Expanded:
image

### DIFF
--- a/frontend/elements/src/entry/jig/_common/sidebar-modules/module.ts
+++ b/frontend/elements/src/entry/jig/_common/sidebar-modules/module.ts
@@ -251,19 +251,14 @@ export class _ extends LitElement {
     }
 
     renderModuleImg() {
-        const { module } = this;
+        const { collapsed, module } = this;
+        
+        if (!collapsed && module === "") return nothing;
 
-        const getIconPath = (module: string) => `entry/jig/modules/small/${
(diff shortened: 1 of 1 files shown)


mastapegs:
closes #1896

frontend/elements/src/entry/jig/_common/sidebar-modules/module.ts
```diff
@@ -250,6 +250,24 @@ export class _ extends LitElement {
         }
     }
 
+    renderModuleImg() {
+        const { module } = this;
+
+        const getIconPath = (module: string) => `entry/jig/modules/small/${
+            module === "" ? "poster" : module
+        }.svg`;
```
MendyBerger, line 258: Why does this have to be a function? why not just make it something like:
```JavaScript
let foo = `/path/${ module || 'empty' }.svg`
```

PR #2118, opened 2021-12-30
feat: remove gray border from iframe

### LINKED ISSUE #1888: Dark gray outlines - Why are you here?
I see these dark gray outlines around the canvas, near the JIG panel, and on top, below the browser panel.
Can we get rid of them?

Screen Shot 2021-11-11 at 14 22 35
k,

mastapegs:
closes #1888 

The gray border was due to the rendered `iframe` element. Removing it required adding the following attribute: `frameBorder="0"`

Image below showing the disappearance of the grey border:
image

### DIFF
--- a/frontend/apps/Cargo.lock
+++ b/frontend/apps/Cargo.lock
@@ -393,6 +393,7 @@ dependencies = [
  "itertools",
  "js-sys",
  "log",
+ "once_cell",
  "serde",
  "serde-wasm-bindgen",
  "serde_json",
--- a/frontend/apps/Cargo.lock
+++ b/frontend/apps/Cargo.lock
@@ -393,7 +393,6 @@ dependencies = [
  "itertools",
  "js-sys",
  "log",
- "once_cell",
  "serde",
  "serde-wasm-bindgen",
  "serde_json",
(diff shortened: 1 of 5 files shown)


MendyBerger (changes requested):
I think that frameborder is deprecated, can you use css for this?

mastapegs:
Yea, definitely deprecated

Screenshot_20211230-075730

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

mastapegs:
Hey there @MendyBerger, I've switched out the use of the deprecated attribute `frameBorder`, with the CSS solution of setting `border: none` to the `iframe`

MendyBerger (changes requested):
This approach to CSS is great for cases where multiple elements need to be styled at once, but in this case there are 2 much simpler ways to go about it:

1. use the dominator `.style` method, works similar to `.property`.
2. If the elements parent is a custom element you can style it with `::slotted`.

mastapegs:
Using `.styles` was SO much easier than applying the `LAZY` CSS. Thank you for that advice @MendyBerger 🙏

And dropped the necessary changes down to a single line.

Issue #2060, opened 2021-12-21
Card games - sound effects for card flips and answers


corinnewo:
I have put the sound effects for the card games that I was given in ...ji-cloud-media>audio>module>cards
https://www.dropbox.com/sh/nqv617gwvfgqc4f/AACjVChAa43UndeCpAkAsBDda?dl=0

There are 3 options for flip - choose one that sounds good. I think the other card sounds are for right and wrong. 

Gali also had CSS effects from https://ianlunn.github.io/Hover/#effects for right and wrong.

Card games: 

- [x] Card quiz
- [x] Memory
- [x] Matching
- [x] Flashcards

Extra:

- [x] Right/wrong css effects

johnnynotsolucky:
@corinnewo the sound effects have been implemented for card-quiz, memory, flashcards and matching games. For each game there are slightly different variations on how the side-effects are played:

#### Card quiz

- Card flip sound is played only when a player makes an incorrect choice;
- If the player makes a correct choice - the positive sound is played.

#### Memory

- Card flip effect for the first card chosen, and then
- Positive or negative effect when the 2nd card is chosen based on the correctness.

#### Flashcards

- Card flip sound is played when a player taps a card and it flips, but not when the card flips back to its original state.
- This game has no right/wrong answers, so no other sound effects were used.

#### Matching

- Positive effect when a card is placed in the correct block;
- Negative effect only if a card is placed in the incorrect block - if the card is dropped outside of the blocks, no effect is played.
  - I made it so that dropping a card outside of a block doesn't play because perhaps the player wanted to change their card and it felt like playing a negative effect would be inappropriate.

johnnynotsolucky:
Animation effect applied to the card-quiz when the correct answer is chosen (#2159)

PR #2134, opened 2021-12-30
feat(fe/components/cards): Implement flip sound effects

johnnynotsolucky:
Part of #2060

- Add a new properties to the play-card element: `eventOnFlipped`;
- Adds an event to the element which dispatches a new `CustomCardFlipped` event if `eventOnFlipped` is true;
- For the card-quiz game, only plays the flipped sound effect if the player makes an incorrect choice
  - Changes in another PR will include a sound effect for making the correct choice.

**Note** doesn't implement flip sound effects for other card games yet.

### DIFF
--- a/frontend/elements/src/module/_groups/cards/play/card/card.ts
+++ b/frontend/elements/src/module/_groups/cards/play/card/card.ts
@@ -95,7 +95,11 @@ export class _ extends LitElement {
     @property({ type: Boolean, reflect: true })
     flipped: boolean = false;
 
-    // whether or not showing front vs. back
+    // whether to dispatch an event when a card is flipped
+    // Note: This is an interim **workaround**. The code that interfaces with this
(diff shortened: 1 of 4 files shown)


frontend/elements/src/module/_groups/cards/play/card/card.ts
```diff
@@ -95,6 +95,10 @@ export class _ extends LitElement {
     @property({ type: Boolean, reflect: true })
     flipped: boolean = false;
 
+    // whether or not showing front vs. back
+    @property({ type: Boolean })
+    eventOnFlipped: boolean = false;
```
johnnynotsolucky, line 100: Using this property on the element itself seemed liked the easiest way without overhauling the `CardOptions` and `Current` structs so that data in those structs could live for `'static`.

The purpose of this property is to prevent the event being dispatched in some cases where we wouldn't want to play the audio effect.

PR #2138, opened 2021-12-31
feat(fe/module/card-quiz): Play positive audio effect for the correct choice

johnnynotsolucky:
Part of #2060 

- Implements `reset_from_jig` and `set_from_jig` for the `AudioMixer` so that there will always be positive and negative clips available;
- Plays a random success effect for the correct choice - Either from the list of configured effects, or from the complete list of available effects if none are configured for the JIG

### DIFF
--- a/frontend/apps/crates/components/src/audio/mixer/mixer.rs
+++ b/frontend/apps/crates/components/src/audio/mixer/mixer.rs
@@ -36,8 +36,27 @@ pub struct AudioSettings {
 }
 
 impl AudioSettings {
-    pub fn reset_from_jig(&mut self, _jig: &JigData) {
-        //TODO...
+    pub fn reset_from_jig(&mut self, jig: &JigData) {
+        // If there is no positive or negative effects configured on the jig, reset the clips to
+        // the list returned by `variants()`. Alternatively we could just **not** reset at all, but
+        // there could potentially be a case where the the audio was reset for a jig with audio effects
+        // and then calling reset_from_jig again would not reset to the defaults.
+
+        self.positive = {
+            if !jig.audio_effects.feedback_positive.is_empty() {
+                jig.audio_effects.feedback_positive.clone().into_iter().collect()
+            } else {
+                AudioFeedbackPositive::variants()
+            }
+        };
+
+        self.negative = {
+            if !jig.audio_effects.feedback_negative.is_empty() {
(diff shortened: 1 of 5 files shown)


frontend/apps/crates/entry/module/card-quiz/play/src/base/state.rs
```diff
@@ -42,6 +43,7 @@ impl Base {
     pub async fn new(init_args: InitFromRawArgs<RawData, Mode, Step>) -> Rc<Self> {
         let InitFromRawArgs {
             jig_id,
+            jig,
```
johnnynotsolucky, line 46: Need access to the jigs data so that we can get the list of audio effects.

frontend/apps/crates/components/src/audio/mixer/mixer.rs
```diff
@@ -36,8 +36,27 @@ pub struct AudioSettings {
 }
 
 impl AudioSettings {
-    pub fn reset_from_jig(&mut self, _jig: &JigData) {
-        //TODO...
+    pub fn reset_from_jig(&mut self, jig: &JigData) {
```
johnnynotsolucky, line 39: This was previously unimplemented. However, I don't think that JIG-related data belongs in the audio_mixer. I'll give this a bit more thought.

johnnynotsolucky: I missed earlier that when a jig is loaded for play/edit, the audio effects are set in the mixer. Before the call to set_from_jig would do nothing, but now it sets the correct sounds.

frontend/apps/crates/components/src/audio/mixer/mixer.rs
```diff
@@ -36,8 +36,27 @@ pub struct AudioSettings {
...
+        // If there is no positive or negative effects configured on the jig, reset the clips to
+        // the list returned by `variants()`. Alternatively we could just **not** reset at all, but
+        // there could potentially be a case where the the audio was reset for a jig with audio effects
+        // and then calling reset_from_jig again would not reset to the defaults.
+
+        self.positive = {
+            if !jig.audio_effects.feedback_positive.is_empty() {
+                jig.audio_effects.feedback_positive.clone().into_iter().collect()
+            } else {
+                AudioFeedbackPositive::variants()
```
johnnynotsolucky, line 49: This might be overly cautious.

PR #2164, opened 2022-01-04
feat(fe/module/edit): Add module mode subtitle for modules which set a mode

### LINKED ISSUE #1564: The name of the use case needs to be a subtitle under the name of the module
It's confusing if you have chosen duplicates as the use case, but when you go into the list, you forget what should be entered into the one list.
Suggest adding a subtitle of the use case under the name of the module on the creation pages.

image

johnnynotsolucky:
Closes #1564 

- Adds a subtitle when editing a module which has modes;
- Shifts the spacing when the subtitle is present so that the content doesn't get squashed too much by the addition of the subtitle element:

#### Without subtitle:

image

#### With subtitle:

Screenshot from 2022-01-04 17-48-26

### DIFF
--- a/frontend/elements/src/module/_common/edit/header.ts
+++ b/frontend/elements/src/module/_common/edit/header.ts
@@ -21,12 +21,27 @@ export class _ extends LitElement {
                     text-align: left;
                     color: var(--dark-blue-4);
                 }
+                :host([subtitle]) .title {
+                    margin-top: 30px;
+                }
+                .subtitle {
+                    margin: 10px 0 10px 0;
+                    font-family: Poppins;
+                    font-size: 24px;
+                    font-weight: 500;
+                    letter-spacing: normal;
+                    text-align: left;
+                    color: var(--dark-blue-4);
+                }
             `,
         ];
     }
 
-    @property()
+    @property({ type: String })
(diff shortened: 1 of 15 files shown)


shared/rust/src/domain/jig/module/body.rs
```diff
@@ -158,6 +158,9 @@ pub trait BodyExt<Mode: ModeExt, Step: StepExt>:
     /// will usually populate an inner .content
     fn new_mode(mode: Mode) -> Self;
 
+    /// Fetch the mode for this module
+    fn mode(&self) -> Option<Mode>;
```
johnnynotsolucky, line 162: Previously there was no consistent way to fetch the mode, only to reset the mode.

frontend/elements/src/module/_common/edit/header.ts
```diff
@@ -21,13 +21,28 @@ export class _ extends LitElement {
                     text-align: left;
                     color: var(--dark-blue-4);
                 }
+                :host([subtitle]) .title {
+                    margin-top: 30px;
```
johnnynotsolucky, line 25: Reduces the top margin when a subtitle is set.

PR #2168, opened 2022-01-04
Refactor/multi persona

### LINKED ISSUE #2070: Registration and Profile: Make persona list multiselect and add in a value

image

RickyLB:
closes #2070 

## What?
Optional multiple personas can be added for `PATCH` and `CREATE` user profiles. 

## How?
`Persona` was changed from type `Option<String>` to `Option<Vec<String>>`.


johnnynotsolucky:
This is great, thanks @RickyLB! Would it make sense to use a Vec instead of Option<Vec> for persona? We wouldn’t need to unwrap or check for None and can operate directly on the list instead.

RickyLB:
@johnnynotsolucky Not a problem to change that to a Vec at all. I'll request a PR review from you so I'm on the same page as frontend. I appreciate the heads up!

PR #2170, opened 2022-01-05
feat(fe/jig/play): move timer ui to the left, away from X

### LINKED ISSUE #2145: Move Timer box In Jig Game so it it not attached to the X that is used to exit a game
When I added a time limit to a Jig, the X that usually appears at the top right of the canvas (used to exit the Jig) lays on top of the timer box, making it appear as if pressing it will deactivate the timer (When in reality it will exit the game)

Suggestion:
Move the timer box so it is detached from the X

Screen Shot 2022-01-02 at 6 22 49 PM
Screen Shot 2022-01-02 at 8 45 54 PM

mastapegs:
closes #2145


mastapegs:
- Don't use inline styles, use Lit-Element
- Move down instead of left

::slotted(slot="main")

mastapegs:
@MendyBerger, I'm now targetting the entire column of indicators, not just the points UI. They're being shifted down, instead of to the left.

MendyBerger:
@mastapegs why positioning? Why not just increase the `padding-top`?

mastapegs:
@MendyBerger That is definitely a simpler solution

PR #2173, opened 2022-01-05
fix(be/multi_persona): remove list with null

### LINKED ISSUE #2070: Registration and Profile: Make persona list multiselect and add in a value

RickyLB:
Part of #2070 

Removed optional vector for list of personas


johnnynotsolucky:
Thanks @RickyLB, I’ll update this PR with some changes for the frontend just now.

johnnynotsolucky (commented):
@RickyLB I made some changes to the backend so that persona is not an Option anywhere. I've included changes to the frontend so it works the same as it did prior - instead now it sends a single persona in a list.

Would you mind reviewing my changes?

I'll make a separate PR to update the UI to support multiple personas.

PR #2204, opened 2022-01-07
feat(fe/module/flashcards): Add card position config

### LINKED ISSUE #2132: Flashcards - Play settings
We need to add in the ability to choose which card is displayed and which is covered.
Ability to choose to show word and cover the picture, or better still, show picture and cover the word.

Same option as in Matching or quiz game, to swap the content. 
How should your cards be displayed?
image

@ChaykeeMor - please add to your list of play settings
@MendyBerger  - please implement - add in a third option

image

johnnynotsolucky:
Closes #2132 

- Adds the card position config option for flashcards game;
- Unused imports cleanup

### DIFF
--- a/shared/rust/src/domain/jig/module/body/flashcards.rs
+++ b/shared/rust/src/domain/jig/module/body/flashcards.rs
@@ -27,6 +27,10 @@ pub struct Content {
 pub struct PlayerSettings {
     /// display mode
     pub display_mode: DisplayMode,
+
+    /// swap the display to be primary left vs. right
+    #[serde(default)]
+    pub swap: bool,
 }
 
 /// Display Mode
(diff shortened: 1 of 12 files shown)


shared/rust/src/domain/jig/module/body/flashcards.rs
```diff
@@ -27,6 +27,10 @@ pub struct Content {
 pub struct PlayerSettings {
     /// display mode
     pub display_mode: DisplayMode,
+
+    /// swap the display to be primary left vs. right
+    #[serde(default)]
```
johnnynotsolucky, line 32: This is a new field so existing modules data will not have it included yet - we need to set a default.

PR #2224, opened 2022-01-10
feat(fe/user): Enable multi-persona selects for user registration and profile

### LINKED ISSUE #2070: Registration and Profile: Make persona list multiselect and add in a value

johnnynotsolucky:
Closes #2070

- Multiple personas can be selected in user profile;
- Multiple personas can be selected during registration.

TODO

- [ ] Add validation to user profile so that at least 1 persona is selected.

### DIFF
--- a/frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
+++ b/frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
@@ -1,6 +1,9 @@
 use super::{actions, state::*};
 use dominator::{clone, html, with_node, Dom};
-use futures_signals::signal::Mutable;
+use futures_signals::{
+    signal::{Mutable, SignalExt},
+    signal_vec::SignalVecExt,
+};
 use std::rc::Rc;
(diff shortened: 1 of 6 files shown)


frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
```diff
@@ -52,22 +55,46 @@ impl Step2Page {
...
-                SimpleSelect::render_mixin(
-                    SimpleSelect::new(
-                        Some(STR_PERSONA_LABEL),
-                        Some(STR_PERSONA_PLACEHOLDER),
-                        None,
-                        STR_PERSONA_OPTIONS.to_vec(),
-                        clone!(state => move |value| {
-                            *state.persona.borrow_mut() = value.map(|x| vec![x.to_string()]);
-                            state.evaluate_persona_error();
+                html!("input-select", {
```
johnnynotsolucky, line 58: This is duplicated logic - I think a good solution is to create a MultiSelect component like the SimpleSelect.

PR #2249, opened 2022-01-12
fix(fe/components): Force dominator to release previous DomHandles

### LINKED ISSUE #2110: After preview mode: Duplicate Jiggling texts appearing on screen
Top left corner still has an empty dialog box. Please remove!

image

johnnynotsolucky:
Resolves #2110 

- Calling `set_inner_html` on an `HtmlElement` does not drop references held for that element. This is because the underlying `DomHandle` that dominator knows about still exists. This change forces dominator to release references held by the previous handle whenever a new signal is fired to render a new body for a page.

### DIFF
--- a/frontend/apps/crates/components/src/module/_common/edit/entry/dom.rs
+++ b/frontend/apps/crates/components/src/module/_common/edit/entry/dom.rs
@@ -1,11 +1,12 @@
 use std::rc::Rc;
 
+use discard::Discard;
 use futures_signals::{
     map_ref,
     signal::{Mutable, SignalExt},
 };
 
-use dominator::{clone, html};
+use dominator::{clone, html, DomHandle};
 
 use utils::{events::ModuleResizeEvent, resize::*};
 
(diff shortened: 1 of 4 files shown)


frontend/apps/crates/components/src/module/_common/edit/entry/dom.rs
```diff
@@ -107,10 +108,19 @@ pub fn render_page_body<Mode, Step, RawData, Base, Main, Sidebar, Header, Footer
...
 
-    state.page_body_switcher.load(sig.for_each(|dom| {
-        let body = dominator::body();
-        body.set_inner_html("");
-        dominator::append_dom(&body, dom);
+    state.page_body_switcher.load(sig.for_each(clone!(state => move |dom| {
+        {
+            // Discard the previous body and set the current handle to None.
+            // This forces dominator to release all references held by this handle.
+            let current_handle = state.dom_body_handle.replace(None);
```
johnnynotsolucky, line 115: `replace()` is necessary because `discard()` moves the value which can't happen if we're holding a reference to it.

frontend/apps/crates/components/src/module/_common/edit/entry/dom.rs
```diff
@@ -107,10 +108,19 @@ pub fn render_page_body<Mode, Step, RawData, Base, Main, Sidebar, Header, Footer
         })
     }));
 
-    state.page_body_switcher.load(sig.for_each(|dom| {
-        let body = dominator::body();
-        body.set_inner_html("");
-        dominator::append_dom(&body, dom);
+    state.page_body_switcher.load(sig.for_each(clone!(state => move |dom| {
+        {
```
johnnynotsolucky, line 112: This scope is not strictly necessary, but it helps visually separate the 2 pieces of logic happening here.

PR #2277, opened 2022-01-14
feat: replace home with iframe of jewishinteractive.org

### LINKED ISSUE #2236: Embed WordPress in iframe on home page
Replace all rows beneath the search panel with the wordpress page https://www.jewishinteractive.org/jigzi-home/ in an iframe

mastapegs:
closes #2236

### DIFF
--- a/frontend/apps/crates/entry/home/src/home/dom/iframe.rs
+++ b/frontend/apps/crates/entry/home/src/home/dom/iframe.rs
@@ -3,9 +3,17 @@ use dominator_helpers::events::Message;
 use futures_signals::signal::{Mutable, SignalExt};
 use std::rc::Rc;
 use web_sys::HtmlIFrameElement;
+use serde::{Serialize, Deserialize};
 
 const STR_IFRAME_URL: &str = "https://www.jewishinteractive.org/jigzi-home";
 const INT_IFRAME_PADDING: usize = 30;
+const INT_INITIAL_HEIGHT: usize = 3000;
+
+#[derive(Serialize, Deserialize, Debug)]
+struct IframeMessageData {
+    kind: String,
+    height: usize,
+}
 
 pub struct Iframe {
     height: Mutable<usize>,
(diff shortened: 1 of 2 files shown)


frontend/apps/crates/entry/home/src/home/dom/mod.rs
```diff
@@ -27,7 +29,7 @@ pub fn render(state: Rc<State>, auto_search: bool) -> Dom {
             html!("empty-fragment", {
                 .child_signal(state.mode.signal_cloned().map(clone!(state => move |mode| {
                     match mode {
-                        HomePageMode::Home => Some(home_sections::render(state.clone())),
+                        HomePageMode::Home => Some(Iframe::render(Iframe::new())), // Some(home_sections::render(state.clone())),
```
mastapegs, line 32: Why not use Iframe::new().render()

mastapegs: Also, use {} and make multiline, so comment is its own line

frontend/apps/crates/entry/home/src/home/dom/iframe.rs
```diff
@@ -0,0 +1,49 @@
...
+        html!("iframe" => HtmlIFrameElement, {
+            .style("width", "100%")
+            .style("border", "none")
+            .style_signal("height", state.height.signal()
+                .map(|height| {
+                    let adjusted_height = height.to_string() + "px";
+                    adjusted_height
+                })
+            )
+            .global_event(clone!(state => move |event: Message| {
```
mastapegs, line 39: .event() doesnt work here. I believe this is because we NEED to listen for the message event on the Window, and not the iframe

MendyBerger: Yeah, looks like you're right

MendyBerger (approved):

Issue #2216, opened 2022-01-10
Shouldn't be possible to preview or publish an empty JIG, or a JIG which has a module with no content


johnnynotsolucky:
**Describe the bug**

A JIG can be published without having any modules added. It can also be published with modules which have no content.

**To Reproduce**

1. Create a new JIG
2. Click Publish
3. Allows you to publish

**Expected behavior**

Shouldn't be able to navigate to the publish screen.

**Additional context**

This should apply to Preview JIG as well - Shouldn't be able to preview a JIG with no modules or modules with no content.

---

- [x] Update related endpoints so that module data includes `is_complete`
- ~[x] In the JIG edit sidebar, change the color of a JIG to red (instead of blue) if there's no content yet;~
  - [x] Update to remove the red color
- [x] Prevent them from publishing until all activities have content set.
  - [x] Update the publish endpoint so that it is not possible to publish when a JIG has incomplete modules
- [x] When previewing a JIG, render a new graphic on modules without content to illustrate that it is waiting for content.
- [x] Update `is_complete()` rules for each game type so that the minimum requirements are met for that type
- [x] On the publish screen notify the teacher that there are modules without content

corinnewo:
@johnnynotsolucky I agree. Can't publish an empty JIG. Should we simply disable the preview button and the publish module if there is no content at all, and if either is clicked, perhaps add in a warning "Wait a minute! Please add some content to your JIG!" @ChaykeeMor - please comment on the warning!

ChaykeeMor:
"Hmm..something's missing. If you'd like to publish this JIG, add activities and a cover."

MendyBerger:
@johnnynotsolucky just keep in mind that we use the player to generate screenshots/previews of the module, so don't disable it on the player level unless we have a solution for screenshots

johnnynotsolucky:
Thanks for the heads-up @MendyBerger!

johnnynotsolucky:
@corinnewo should it be possible to publish _without_ a cover?

johnnynotsolucky:
Disabling the preview/publish buttons when there are no modules is simple enough, but it is currently not possible to know if all the modules for a JIG have content. The data returned by `/jig/<id>/draft` endpoint includes the module ID and type of module. 

@MendyBerger I wonder if we should have an endpoint we can call to tell us whether a jig can be previewed or published? It can be called whenever the author clicks preview/publish. 

[...]

PR #2309, opened 2022-01-20
feat(fe/components/cards): Implement minimum requirements for card games

johnnynotsolucky:
Part of #2216

- Implements rules for minimum requirements in card games:
  - Words and Images: at least 2 pairs (max of 14), text card must not have empty text and images must be set on both pairs;
  - All other card types: at least 2 pairs (max of 14) and cards may not have empty text.
- Update so that if a teacher wants to replace the word list, the state of the module is marked as incomplete (previously it would only mark as complete once and never update again);
- Moves MAX_LIST_WORDS to shared config; Adds MIN_LIST_WORDS to shared config (previously it was a magic number `2`).

Extra work included in this PR:
- Fixed dual and single lists so that their validation logic catches mismatched lists, empty text, etc.

### DIFF
--- a/frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_1/state.rs
+++ b/frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_1/state.rs
@@ -19,7 +19,10 @@ use crate::{
 use dominator::clone;
 use futures_signals::signal::Mutable;
 use once_cell::sync::OnceCell;
-use shared::domain::jig::module::body::{Image, _groups::cards::Mode};
+use shared::{
+    domain::jig::module::body::{Image, _groups::cards::Mode},
+    config as shared_config,
+};
 use utils::unwrap::UnwrapJiExt;
 use std::rc::Rc;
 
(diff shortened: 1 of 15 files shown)


frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_1/state.rs
```diff
@@ -139,8 +142,8 @@ fn make_single_list<RawData: RawDataExt, E: ExtraExt>(
     );
 
     let options = SingleListOptions {
-        max_rows: config::MAX_LIST_WORDS,
-        min_valid: 2,
```
johnnynotsolucky, line 143: Removes magic number

frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_1/state.rs
```diff
@@ -165,14 +168,14 @@ fn make_dual_list<RawData: RawDataExt, E: ExtraExt>(
...
     let options = DualListOptions {
-        max_rows: config::MAX_LIST_WORDS,
+        max_rows: shared_config::MAX_LIST_WORDS,
         cell_rows: {
             match state.base.mode {
                 Mode::Riddles => 2,
                 _ => 1,
             }
         },
-        min_valid: 2,
```
johnnynotsolucky, line 175: Same https://github.com/ji-devs/ji-cloud/pull/2309/files#r788528732

PR #2317, opened 2022-01-21
Incomplete error tooltip

### LINKED ISSUE #2216: Shouldn't be possible to preview or publish an empty JIG, or a JIG which has a module with no content

johnnynotsolucky:
Closes #2216 

- Adds error tooltip to the first incomplete module in a JIG when the teacher clicks publish but has incomplete modules;
- Highlights incomplete modules in red;
- Scrolls to first incomplete module when clicking publish;
- Adds error tooltip to Publish window in the sidebar if the teacher clicks publish without adding any activities or a cover;
- Disables the publish button on the publish screen if the teacher hasn't added any activities or a cover.

Outstanding:
- [ ] ~Work out how to get the left-pointing arrow to show up;~
  - Currently only the up and down arrows show up when tracking the position of the module in the sidebar.
  - I have a draft PR with a POC refactor of tooltips to make them more practical: #2322 
- [x] Close the tooltip when the teacher clicks anywhere (or worst-case, just on the tooltip itself).

### Current progress:

image

### DIFF
--- a/frontend/apps/crates/entry/jig/edit/src/edit/publish/dom/mod.rs
+++ b/frontend/apps/crates/entry/jig/edit/src/edit/publish/dom/mod.rs
@@ -240,7 +240,12 @@ fn render_page(state: Rc<Publish>) -> Dom {
                     .child(html!("button-rect", {
                         .text(STR_PUBLISH)
                         .text(state.jig.focus_display())
-                        .property("disabled", state.jig.modules.lock_ref().iter().find(|m| !m.is_complete).is_some())
+                        .property(
+                            "disabled",
+                            state.jig.jig_focus.is_modules()
+                            && (state.jig.modules.lock_ref().len() == 0 // Disabled
+                                || state.jig.modules.lock_ref().iter().find(|m| !m.is_complete).is_some())
+                        )
                         .child(html!("fa-icon", {
                             .property("icon", "fa-light fa-rocket-launch")
                             .style("color", "var(--main-yellow)")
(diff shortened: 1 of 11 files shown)


frontend/apps/crates/entry/jig/edit/src/edit/publish/dom/mod.rs
```diff
@@ -240,7 +240,12 @@ fn render_page(state: Rc<Publish>) -> Dom {
                     .child(html!("button-rect", {
                         .text(STR_PUBLISH)
                         .text(state.jig.focus_display())
-                        .property("disabled", state.jig.modules.lock_ref().iter().find(|m| !m.is_complete).is_some())
+                        .property(
```
johnnynotsolucky, line 243: Even though the teacher won't be able to navigate to the publish screen via the sidebar, if they know to add /publish to the url, they'll see the screen. This is a safety check.

PR #2330, opened 2022-01-25
MAJOR: Modules now have ThemeId instead of ThemeChoice

### LINKED ISSUE #2238: Theme component - combined with former background component
All flow link: https://zpl.io/a8dpd4Q

See all the changes:
*** In the Zeplin the texts are final, (on the screenshots they are not updated with last texts changes)

1. On all activities (except the card's activities) - the background step (1) is now replaced with the updated theme component. 
The idea is that the teacher always starts with the theme panel, and only if they want to, they open the background design flow. Its not the next step in the flow after the theme selection anymore as this was confusing. I show the design on the cover, but this will be consistent in all the activities (except the cards games).
Screen Shot 2022-01-11 at 12 33 02

2. Take out the yellow mark for JIG theme. Keep only the selection for the local activity theme. Take out the popup to set as JIG theme. We make it back to simplicity (sorry...). What the teacher selected in this activity will be the default selection in future activities. If on the next activity the teacher changes the theme - change is only for that activity. On the JIG settings, if the teacher goes there and selects a theme for the JIG, this is the default theme she sees on each new activity. She can change locally. If she changes later in the JIG settings, it will change for future activities. not changing the former ones. If the teacher wants to change former activities she has to make it on each activity. @corinnewo What do you think about this section??
1920: https://zpl.io/bW5Xxmj
1280: https://zpl.io/VxjrZQJ
Screen Shot 2022-01-11 at 12 35 50

3. Add a hyperlink icon "Design from scratch" (hover is a darker blue icon and text)
Screen Shot 2022-01-11 at 12 40 17

4. On click on this button a popup appears with 2 tabs: background and overlays.
1920: https://zpl.io/V0ApBjm
1280: https://zpl.io/2GM10wj
Screen Shot 2022-01-11 at 12 44 17

5. Fill color is not a tab, the same reason, it was confusing as a flow. So it is a hyperlink button option on the background tab. 
Screen Shot 2022-01-11 at 12 46 04

6. On click on the "Select fill color" button, a colors popup appears. 
I made the popup elements smaller so the same size of the popup is used for big and small screen sizes. Pls make sure to reduce all sizes on this component. I also created some rules of where to place the popup, maximum height, margin from top and bottom of the page for 1920 and 1280 screen sizes. Pls let me know if all the rules are clear, we can meet to discuss this. The same behavior is used for this colors popup on the text component. 
1920: https://zpl.io/brRPZXD, https://zpl.io/2ZdXN7l
1280: https://zpl.io/255xB8J, https://zpl.io/bJB1R4J, https://zpl.io/aXpXv8x
Screen Shot 2022-01-11 at 12 56 31
Screen Shot 2022-01-11 at 12 56 47
Screen Shot 2022-01-11 at 12 57 39
Screen Shot 2022-01-11 at 12 58 04

7. On click on the Add color button, a picker popup appears on top of the colors popup.
I made the picker popup much smaller. Pls add the "OK" button to this picker popup. See instructions for placing the picker popup.
Same size for all screens sizes: https://zpl.io/V1gDBBJ
@MendyBerger Can we add an icon that picks a color from an image?
Screen Shot 2022-01-11 at 14 12 15

MendyBerger:
#### The old way:
Modules can either use the jigs theme (default), or have their own theme, the `ThemeChoice` enum was used for this.

#### The new way:
Modules all have their own themes, when a new module is created it takes the jigs theme, and when a module changes theme that changes the jigs theme.
The jig's theme doesn't have much use, it's only needed to get the default theme for modules.
Since the jig's theme is almost never used the theme editor on the jig level will be removed, the only way the jig's theme can be changed is by selecting a theme on one of the modules and that will propagate to the jig.

At this point there is only one commit, albeit a large one, and that removes the `ThemeChoice` enum.

@dakom @johnnynotsolucky I'd really appreciate if you can both take a look at the changes I made, since this is a major change to a section of the codebase that I'm not super familiar with.

@RickyLB will have to do migrations on the old modules once this is merged.

resolves #2238


dakom:
@MendyBerger as you noted this is a pretty big change with a lot of moving parts... if the compilation tests pass, then my next step would be to really sit through and try it out from a user experience perspective, which is probably better for QA?

If there's any specific area you're concerned about, I could speak to that though... or we could jump on zoom and you could walk me through a specific pipeline you're concerned about.

@corinnewo fwiw I think this decision makes a lot of sense and will make things easier to maintain. Glad to hear it was approved.

frontend/apps/crates/components/src/module/_groups/cards/edit/sidebar/step_2/actions.rs
```diff
@@ -1,10 +1,10 @@
-use shared::domain::jig::module::body::{Background, ThemeChoice};
+use shared::domain::jig::module::body::{Background, ThemeId};
 
 use crate::module::_groups::cards::edit::state::*;
 
 impl<RawData: RawDataExt, E: ExtraExt> CardsBase<RawData, E> {
-    pub fn set_theme(&self, theme: ThemeChoice) {
-        self.theme_choice.set_neq(theme);
+    pub fn set_theme(&self, theme: ThemeId) {
```
johnnynotsolucky, line 6: In #2337 I moved the separate implementations of `CardBase<> ` into "_groups/cards/edit/actions.rs" so that some of the functionality was accessible outside of individual steps - instead of having a super::actions, step_1::actions, etc, I merged them all together.

frontend/apps/crates/components/src/module/_groups/cards/edit/state.rs
```diff
@@ -11,7 +11,7 @@ use futures_signals::{
 use shared::domain::jig::{
     module::{
         body::{
-            Background, BodyExt, Image, Instructions, ThemeChoice,
+            Background, BodyExt, Image, Instructions,
             _groups::cards::{self as raw, BaseContent, Mode, Step},
```
johnnynotsolucky, line 15: To resolve the conflict on this file:

```suggestion
            Background, BodyExt, Image, Instructions,
            _groups::cards::{self as raw, BaseContent, Mode, Step}, Audio
```

frontend/apps/crates/entry/module/card-quiz/edit/src/debug.rs
```diff
@@ -6,7 +6,7 @@ use once_cell::sync::OnceCell;
 use shared::domain::jig::{
     module::{
         body::{
-            Instructions, ThemeChoice,
+            Instructions,
             _groups::cards::{BaseContent, Card as RawCard, CardPair as RawCardPair, Mode, Step},
```
johnnynotsolucky, line 10: To resolve the conflict in the debug.rs files:

```suggestion
            Instructions,
            _groups::cards::{BaseContent, Card as RawCard, CardContent as RawCardContent, CardPair as RawCardPair, Mode, Step},
```

PR #2337, opened 2022-01-26
Adds context menu and audio widget to card games when editing

johnnynotsolucky:
Part of #2280

- Adds audio tab with audio widget
  - Widget will only active once a card or pair is selected (not yet implemented)
- Updates menu kebab so that it can accept a custom container which can be placed with OverlayHandle for correct positioning
- Updates theme properties for cards so that they're set in css instead of as a style on individual html elements;
  - Makes it easier and neater to update them when necessary.
- Card/set selections;
- Per-card context menu to remove audio, images, pairs;
  - **Note:** Not confirm modals yet.
- Per-card hover state with lighter border for default state;
- Updates Card data type so that Audio can be stored in the database.
  - Includes `Deserialize` implementation for `Card` so that existing data based on the original `Card` enum can be correctly mapped to the new Card data.
- Record audio for a card or set of cards
  - Selecting a set where each card has different audio files will mean that the teacher must record a new audio file
  - **Note:** Audio icon appears but non-functional still

image


corinnewo:
Looks impressive! Just an aside - there may be an image tab as well, not just text tab (in use case where there is text and image)

johnnynotsolucky:
@corinnewo yip, I should’ve used the Words and Images use case for the screenshot. It does work correctly with the Image tab. 

There is a problem with the jiggling text, it assumes the theres _always_ an image tab thats 2nd, so whenever the audio tab is navigated to the Image tab jiggling text shows.

corinnewo:
The jiggling text is problematic- David set up a json file system, which is based on tab position only. So here, it worked that if there was the extra images tab, the text was correct. Now, where sometimes there is the extra images tab, but always an audio tab, there is a problem! How can we deal with this in the json? If we always put the audio tab second, it could work. What do you suggest?

Sent from my iPhone

Corinne Ossendryver
CPO, Jewish Interactive 
(+27) 079-8865326

johnnynotsolucky:
@corinnewo I've created #2341 to handle the JSON file and Jiggling text.

frontend/apps/crates/components/src/module/_groups/cards/edit/actions.rs
```diff
@@ -0,0 +1,111 @@
...
+    if len > max_len {
+        let cutoff_grapheme_byte = text.grapheme_indices(true).nth(max_len).unwrap_ji().0;
+
+        text[..cutoff_grapheme_byte].to_string()
+    } else {
+        text
+    }
+}
+
+impl<RawData: RawDataExt, E: ExtraExt> CardsBase<RawData, E> {
```
johnnynotsolucky, line 19: Moved this implementation higher up the hierarchy so that it's accessible by all callers.

frontend/apps/crates/components/src/module/_groups/cards/edit/actions.rs
```diff
@@ -0,0 +1,111 @@
...
+                    .map(|pair| RawCardPair(pair.0.into(), pair.1.into()))
+                    .collect();
+            }
+        });
+    }
+
+    /// Replace a pair at `idx` with the pair returned by `f`
+    ///
+    /// `f` is passed the pair originally at `idx` in the list of pairs.
+    pub fn replace_pair<F>(&self, idx: usize, f: F)
```
johnnynotsolucky, line 74: New function to update a specific pair in the list.

frontend/elements/src/_themes/themes.ts
```diff
@@ -132,6 +144,10 @@ function setRootVars() {
                 `--theme-${id}-cards-border-color-var`,
                 rgb(value.borderColor, "var")
             );
+            style.setProperty(
+                `--theme-${id}-cards-border-color-light-hsl`,
```
johnnynotsolucky, line 148: New custom property so that we can have a lighter default border around cards.

frontend/elements/src/core/menu/kebab.ts
```diff
@@ -41,30 +41,68 @@ export class _ extends LitElement {
...
+    @property({ type: Boolean })
+    positioningEnabled: boolean = true;
+
+    /// Whenever this element needs to be used with some other container outside of this element,
+    /// the user can set this property so that clicking in it doesn't dispatch any events.
+    @property()
+    customContainer?: HTMLElement;
+
+    /// Set the event listener so that it can be reliably removed later.
+    eventListener: any;
```
johnnynotsolucky, line 57: Important additions so that users of this element can pass in custom containers, notably anything rendered with `OverlayHandle`.

frontend/ts-utils/hex.ts
```diff
@@ -44,3 +44,48 @@ export function hexRgbToNumber(r: number, g: number, b: number): number {
 export function hexNumberToString(n: number, prefix: string = "#"): string {
     return prefix + n.toString(16);
 }
+
+export function rgbToHsl(r: number, g: number, b: number): [number, number, number] {
```
johnnynotsolucky, line 48: https://css-tricks.com/converting-color-spaces-in-javascript/

shared/rust/src/domain/jig/module/body/_groups/cards.rs
```diff
@@ -63,9 +66,117 @@ pub struct EditorState {
...
+//   audio: null,
+//   card_content: {"Text": "Some words"}
+// }, {
+//   audio: null,
+//   card_content: {"Image": {<Image data>}}
+// }]
+//
+// TODO Create a content migration to migrate all existing JIGs with card game modules so that
+// their card data matches the new Card struct and delete this Deserialize implementation.
+impl<'de> de::Deserialize<'de> for Card {
```
johnnynotsolucky, line 98: NB, required so we don't break existing jigs.

PR #2372, opened 2022-02-04
feat(be/age_range): add short display name to AgeRange

RickyLB:
Added optional `short_display_name` to `AgeRange` object and added the field to the database.


RickyLB:
@mastapegs

PR #2380, opened 2022-02-06
feat: change jig card age text

### LINKED ISSUE #2055: Creator menu - change ages names + Add to action menu play and share options
@meravs 

**Creator menu_01_MS**
(As we have later versions which were not applied at this stage, Can you name the new file "Creator menu old version_02_MS"?)

1. Ages are changed following the search results shortcuts and presenting range.
As in issue #2033  (no. 3).
Screen Shot 2021-12-21 at 10 44 21
Screen Shot 2021-12-21 at 10 44 27

2. Play and share options are added to the actions menu of each JIG.
Screen Shot 2021-12-21 at 10 41 21

Pls upload here the new screenshots and the Zeplin links, including the components page of the new icons.

mastapegs:
closes #2055

### DIFF
--- a/frontend/apps/crates/utils/src/ages.rs
+++ b/frontend/apps/crates/utils/src/ages.rs
@@ -1,25 +1,43 @@
 pub use shared::domain::jig::module::body::ThemeId;
 use shared::domain::meta::{AgeRange, AgeRangeId};
 
+const STR_ALL_AGES: &str = "All Ages";
+const STR_DASH: &str = "-";
+
 pub trait AgeRangeVecExt {
     fn range_string(&self, selected: &Vec<AgeRangeId>) -> String;
 }
 
 impl AgeRangeVecExt for Vec<AgeRange> {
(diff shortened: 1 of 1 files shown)


frontend/apps/crates/utils/src/ages.rs
```diff
@@ -7,19 +7,29 @@ pub trait AgeRangeVecExt {
...
+                .iter()
+                .rev()
+                .find(|age| selected.contains(&age.id))
+                .unwrap();
 
-        let mut result = String::new();
-        if let Some(first_age) = first_age {
-            result = first_age.display_name.clone();
-        };
+            let first_age_text = first_age.short_display_name.as_ref().unwrap_or(&first_age.display_name);
```
mastapegs, line 23: `short_display_name` is an `Option`. Using `display_name` as fallback if `short_display_name` is None.

MendyBerger:
Like the solution with reverse iteration, cool idea!

Just two comments
1. Don't think the case where only one age is selected is handled.
2. Should also say "All ages" if no ages are selected

PR #2462, opened 2022-02-25
feat: add jig settings icon hover state

### LINKED ISSUE #2431: Create: Jig Kebab panel - correct icon?
I've noticed that the settings icon by the kebab menu panel looks different - @GaliShapira is this the correct design?
 
**Currently on Jigzi:**
Screen Shot 2022-02-20 at 9 52 49 PM

**Previous icon:**
Screen Shot 2022-02-20 at 9 52 29 PM

mastapegs:
closes #2431 

This PR:

- Extracts the Jig Settings Button into its own component
- The component holds onto the hover state and captures how that state can change in the update fn

The issue is that Font Awesome doesn't have a light version of their gear icon that matches the the desired ticket. Here are the Normal vs Hover states of the icon bundled in with the PR currently:

Normal:
image

Hover:
image

@MendyBerger, should I copy the images from #2431 and use those for the gear icons? Or is there some way we can still use Font Awesome's gear, but make it lighter?


mastapegs:
I'm going to amend this PR

Instead of changing the icon, we're going to make the BG color lighter.

That blue color behind the gear icon is defined here:
https://github.com/ji-devs/ji-cloud/blob/4c55f4a3114deb17d054f89f38f1a98d1522cdc3/frontend/elements/src/entry/jig/edit/sidebar/header.ts#L100

The goal is to change the `background-color` on the `anchored-overlay` from `--main-blue` to `--light-blue-6`

PR #2491, opened 2022-03-03
feat: change blocked jig state from curation table

mastapegs:


frontend/apps/crates/entry/admin/src/curation/table/dom.rs
```diff
@@ -74,77 +83,123 @@ impl CurationTable {
...
-                                    .text_signal(state.age_label(age_id))
-                                })
-                            }))
+                            // .child_signal(jig.age_ranges.signal_cloned().map(|ages_hash| {
+                            //     ages_hash.into_iter().map(|age_id| {
+                            //         html!("span", {
+                            //             .text(state.age_label(age_id))
+                            //         })
+                            //     })
+                            // }))
```
mastapegs, line 187: I can't seem to pass Mutable HashSets in as children. This may be an easy issue to fix

MendyBerger:
Nice job @mastapegs!!

Just one comment, looks like you forgot to lift up `EditableJig`, it's still in `/curation/jig` even though it's also used in `/curation/table`, but merging anyway since the code seems to be working correctly so you can do this is a separate pr

PR #2496, opened 2022-03-04
feat: update the beta popup text

### LINKED ISSUE #2488: Beta version popup - needs more explanation
@ChaykeeMor Need your help here, please!

At the moment we have this on our home page: 
image

Sammy raised a point yesterday that people may not know what BETA means.  Opening up the popup does not explain anything further!

We thought of maybe changing the heading of the yellow popup to something like "You are using the first version of Jigzi" and to add something like "We couldn't wait for you to start using Jigzi! This is not a full version yet - we have a lot exciting modules more planned for you! Please be patient, and tell us what you think so far!"

Probably needs to be much shorter, but that is the main idea - to explain that Beta means it is the first release.... Thanks

mastapegs:
closes #2488

image

### DIFF
--- a/frontend/apps/crates/components/src/page_header/state.rs
+++ b/frontend/apps/crates/components/src/page_header/state.rs
@@ -18,7 +18,7 @@ impl State {
         Self {
             logged_in: Mutable::new(LoggedInState::Loading),
             loader: AsyncLoader::new(),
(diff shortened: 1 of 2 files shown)


mastapegs:
This PR updates:

- Beta popup text
- Initial popup state. Was hidden by default initially. now its active initially

PR #2505, opened 2022-03-07
feat: dont unnecessarily show hubspot cookie

### LINKED ISSUE #2464: Embedded JIGs - cancel the cookie
When you embed a game into another site, then playing the game brings the cookie question each time. We do not want the child to have to accept cookies or not. Can we cancel the question in embedded JIGs?
I created an example embed here: https://www.jewishinteractive.org/embed-test/

image

mastapegs:
closes #2464

### DIFF
--- a/backend/pages/src/templates/spa.rs
+++ b/backend/pages/src/templates/spa.rs
@@ -62,6 +62,7 @@ struct SpaPageInfo {
     firebase: bool,
     google_maps_url: Option<String>,
     local_dev: bool,
+    include_hubspot: bool,
 }
 
 fn spa_template(settings: &RuntimeSettings, spa: SpaPage) -> actix_web::Result<HttpResponse> {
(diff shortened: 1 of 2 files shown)


mastapegs:
@MendyBerger 2 things:
- I wanted to make sure that I'm excluding enough routes from showing the hubspot code. Right now, I'm excluding:
  - Jig
  - Module
  - Legacy Jigs
- `rust-analyzer` REALLY wanted to update the `Cargo.toml`. I did not add any deps, but it seems that rust-analyzer wanted to add semvar info to some packages. Should I leave the `Cargo.toml` changes in?

MendyBerger:
@mastapegs sorry it took me so long to respond.

Excluded routes are good with the exception of jig/edit, only jig/play should not get hubspot, jig/edit does need to have it.

I think you should remove the lock file changes since you didn't touch the dependencies, and if I understand correctly the lock file is there just to keep track of the dependencies

MendyBerger:
@mastapegs only the jig route should have hubspot on edit, the module should never have it

PR #2564, opened 2022-03-17
feat(fe/module/memory): Add pairs per game activity setting

### LINKED ISSUE #2426: Memory Game - add setting
We are no longer limiting the amount of pairs allowed for card games but memory games can only fit up to 14 pairs. We therefore need to add a setting to the memory game that allows users to decide how many pairs to include [at random] in their game. rollover on setting: Display [dropdown 4-14] pairs. 

Screen Shot 2022-02-17 at 16 30 01

Not designed yet, but we will also have a warning when copying list-set from another module with more than 14-- receive warning "Memory games can only fit 14 pairs on a page. What would you like to do? randomly generate memory game from pairs/split into multiple memory games/cancel memory game"

@corinnewo just dropping this here so it doesn't get lost. Will come back to this when we implement image-image cards, right?

johnnynotsolucky:
Closes #2426

- Adds "pairs per game" setting to memory game activity
	- Single setting icon which can be disabled for default system defined pairs per game, or enabled and user-set;
	- When disabled, uses maximum (14) pairs _or_ total pairs if that total is less than the maximum of 14;
	- When enabled, a teacher can set the value to between 2 and 14 pairs, inclusive. Entering a value of less than 2 will reset it to 2. Entering a value of more than 14 will reset it to 14.
- Fixes number circle value
	- Removes SVG for rendering the circle + number, uses a regular `<span />` and css.
	- The SVG was causing number values > 10 to render to the right, not centered.
- Fixes an issue where clicking inside the setting bubble triggered a click event on the setting icon
- Prevent setting bubble from showing up when hovering over inactive/disabled setting icons

#### Randomized pairs when total count of pairs is more than the setting

_Uses the alphabet for testing `[a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q]`_

_Note: temporarily flipped all cards so it's easier to visually see that things are working correctly_

Screenshot 2022-03-17 at 13 40 57
Screenshot 2022-03-17 at 13 41 15
Screenshot 2022-03-17 at 13 41 32

#### Enabled, user set
Screenshot 2022-03-17 at 13 42 01

#### Default to total pairs if less than max
Screenshot 2022-03-17 at 13 43 25

#### Defaults to max pairs if total pairs more than max
Screenshot 2022-03-17 at 13 41 49

#### Setting set to 7 pairs
Screenshot 2022-03-17 at 13 42 18

### DIFF
--- a/frontend/apps/crates/entry/module/memory/play/src/base/state.rs
+++ b/frontend/apps/crates/entry/module/memory/play/src/base/state.rs
@@ -1,4 +1,4 @@
-use shared::domain::jig::{
+use shared::{domain::jig::{
     module::{
         body::{
             Background, Instructions,
@@ -8,7 +8,7 @@ use shared::domain::jig::{
         ModuleId,
     },
     JigId,
-};
+}, config::MAX_LIST_WORDS};
 
 use super::card::state::*;
 use components::module::{_common::play::prelude::*, _groups::cards::lookup::Side};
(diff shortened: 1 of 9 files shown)


frontend/apps/crates/entry/module/memory/play/src/base/state.rs
```diff
@@ -53,18 +53,36 @@ impl Base {
...
+                    pairs_to_display as usize
+                }
+            });
+
+        let n_cards = max_cards * 2;
         let mut pair_lookup: Vec<usize> = vec![0; n_cards];
         let mut cards = {
-            let pairs = &content.base.pairs;
+            let pairs = &mut content.base.pairs;
+            pairs.shuffle(&mut rng);
```
johnnynotsolucky, line 80: Shuffle the pair so that if there are more pairs than the max or than specified in the setting, each time a student plays the game, a different set of pairs should be rendered

PR #2568, opened 2022-03-21
feat(fe/user): adjustments to user step 2

### LINKED ISSUE #2556: Sign up - Step 2 - terms and conditions not aligned with the other text
Why is terms and conditions about 3 pixels higher than the rest of the sentence? Please align.
image

Also, please break the sentence above 'one more step' into smaller parts so that it is not one long sentence across the screen. Perhaps  set margin at the end of the input fields?

mastapegs:
closes #2556 

image


frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
```diff
@@ -112,41 +112,36 @@ impl Step2Page {
...
+                                    .style("color", "var(--main-blue)")
+                                }))
+                                .child(html!("span", {
+                                    .text(STR_TERMS_LABEL_TWO)
+                                }))
+                                .child(html!("a", {
+                                    .attribute("target", "_blank")
+                                    .attribute("href", "https://www.jewishinteractive.org/privacy-policy/")
+                                    .text(STR_TERMS_PRIVACY)
+                                    .style("color", "var(--main-blue)")
```
mastapegs, line 135: same example as previous comment

frontend/apps/crates/entry/user/src/register/pages/step_2/dom.rs
```diff
@@ -24,6 +25,10 @@ pub struct Step2Page {}
 
 impl Step2Page {
     pub fn render(step: Mutable<Step>, step_1: Step1Data) -> Dom {
+        static LINK_COLOR: Lazy<String> = Lazy::new(|| class! {
+            .style("color", "var(--main-blue)")
+        });
+
```
mastapegs, line 31: Hey @MendyBerger, what's your opinion of using styles here?

Moving the anchor styles over to the Web Component side for these anchor tags feels very specific for logic that lives in the `input-checkbox` component. I moved the other styles into `input-checkbox` but should I leave the anchor colors in Rust?

MendyBerger: Hmm..., I think we should chat about this face to face, gimme a ping when you're available

frontend/elements/src/core/inputs/composed/checkbox.ts
```diff
@@ -83,7 +86,9 @@ export class _ extends LitElement {
                             .checked=${checked}
                             @change="${this.onChange}"
                         />
-                        <span class=""> ${label} </span>
+                        <span class="">
+                            ${label} <slot name="label"></slot>
```
MendyBerger, line 90: Can you put the `${label}` in the slot?

MendyBerger (changes requested):

PR #2588, opened 2022-03-29
optimize(be): Add browse jig count tracing calls

johnnynotsolucky:
- Adds rust-toolchain.toml to backend crate;
- Adds query instrumentation to browse count queries.

### DIFF
--- a/backend/api/rust-toolchain.toml
+++ b/backend/api/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "stable"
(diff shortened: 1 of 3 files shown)


backend/api/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "stable"
```
johnnynotsolucky, line 2: @RickyLB the backend builds in CI with stable afaik, I've set this here so that I don't need to switch toolchains manually.

PR #2615, opened 2022-03-31
feat: refactor module routers

mastapegs:
@MendyBerger and @johnnynotsolucky , while working in the module area, I've noticed that the routing logic used seems a bit verbose. I wanted to ask your opinion on these changes

### DIFF
--- a/frontend/apps/crates/entry/module/card-quiz/edit/src/router.rs
+++ b/frontend/apps/crates/entry/module/card-quiz/edit/src/router.rs
@@ -28,24 +28,11 @@ pub fn render(state: Rc<Router>) {
         dominator::routing::url()
             .signal_ref(|url| Route::from_url(url))
             .for_each(clone!(state => move |route| {
-                match route {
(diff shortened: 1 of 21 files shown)


frontend/apps/crates/entry/module/card-quiz/edit/src/router.rs
```diff
@@ -28,24 +28,11 @@ pub fn render(state: Rc<Router>) {
...
-                                    }
-                                    _ => {}
-                                }
-                            }
-                            _ => {}
-                        }
-                    },
-                    _ => {}
-                };
+                if let Route::Module(ModuleRoute::Edit(ModuleKind::CardQuiz, jig_id, module_id)) = route {
```
mastapegs, line 31: I just think that this line reads easier

"If the `Module` we are `Edit`ing is a `CardQuiz`, then"

MendyBerger:
Nice job @mastapegs! Clippy is going to be really happy!

PR #2628, opened 2022-04-05
feat: set student to true for embed code

### LINKED ISSUE #2561: Embedding issues
There are a few embedding issues when you copy and paste the embed code into another website:

1. The embed code has the jig on the left, which is aimed at teachers, which is fine, EXCEPT that when you try and share a code, the embed crashes. I think we should embed using the student player (no JIG panel), so that there are no such issues. I also think that this makes sense, as most embedded JIGs will be for students to play. [**is_student=true** in the embed code]
2. On the last screen, the popup should have some cell space at the top and bottom. It is currently touching the top and bottom. 
3. At the end of the module, there are 3 options: 
-**replay**, which works
-**share**, which causes a **crash** if you share a code, copying the link does not work either 
   NOTE: IF THIS IS A STUDENT PLAYER, THIS WILL NOT HAPPEN
-**exit**, which does nothing

Is it possible to have a different end screen when the game is embedded? Possibly with only the replay function? The student player has an exit as well at the end, which won't work. Is it possible to hide the exit and only have the replay?

image

mastapegs:
closes #2561 

# Before

image

---------------------------------------

# After

image

### DIFF
--- a/frontend/apps/crates/components/src/share_jig/state.rs
+++ b/frontend/apps/crates/components/src/share_jig/state.rs
@@ -41,7 +41,10 @@ impl ShareJig {
         let url = Route::Jig(JigRoute::Play(
             self.jig_id,
             None,
-            JigPlayerOptions::default(),
+            JigPlayerOptions {
+                is_student: true,
+                ..Default::default()
+            },
         ))
         .to_string();
         let origin = web_sys::window()


MendyBerger:
@mastapegs I wouldn't change the from impl, can you do it specifically for the embed code?

Issue #2651, opened 2022-04-11
Add SHARE link to the search results card


corinnewo:
Connected with #2258 
Add in the **Share** and **Edit** links to the cards in the search results. 
1. Share - opens up the same popup the opens up from within the player.
2. Edit (only visible if it is a JIG that I created) - opens the JIG up within the creator, ready to edit.

![image](https://user-images.githubusercontent.com/16290361/162749034-1e7275da-c87c-4f3e-9cba-736dd5584dbc.png)

corinnewo:
@mastapegs Also, while you working on the search result cards, please change the - to > as per #2607, and fix up the age range.

johnnynotsolucky:
@GaliShapira would you be able to send me the play button arrow for this task?

GaliShapira:
@johnnynotsolucky 
See attached the icon. Pls make sure to have a hover color for the button as in the Zeplin.
![icn-play](https://user-images.githubusercontent.com/88834002/167394755-9001f9b5-9224-4df9-b67e-327970eb6d31.svg)
https://zpl.io/2jOl18r
<img width="331" alt="Screen Shot 2022-05-09 at 13 44 36" src="https://user-images.githubusercontent.com/88834002/167394762-156e93bf-0413-43d2-bb1f-ceb863499628.png">

PR #2719, opened 2022-05-10
feat(fe/jig/search): Add jiggling on back of JIG search results

johnnynotsolucky:
Part of #2651 

- Adds a Jiggling to the back of JIG search results cards

https://user-images.githubusercontent.com/4161106/167582364-9340639c-e87c-47fc-957a-3fe09460ac05.mov

### DIFF
--- a/frontend/elements/src/entry/home/home/search-results/search-result.ts
+++ b/frontend/elements/src/entry/home/home/search-results/search-result.ts
@@ -15,6 +15,30 @@ export class _ extends LitElement {
     static get styles() {
         return [
             css`
+                @keyframes jump {
+                    0% {
+                        transform: scaleX(-1) translateY(0px);
+                    }
+                    20% {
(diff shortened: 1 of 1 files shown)


frontend/elements/src/entry/home/home/search-results/search-result.ts
```diff
@@ -15,6 +15,30 @@ export class _ extends LitElement {
     static get styles() {
         return [
             css`
+                @keyframes jump {
+                    0% {
+                        transform: scaleX(-1) translateY(0px);
+                    }
+                    20% {
+                        transform: scaleX(-1) translateY(0px);
+                    }
```
johnnynotsolucky, line 24: We need to wait a bit for the card flip animation to complete before animating the jiggling into view

johnnynotsolucky:
This doesn't work correctly on safari.

Issue #2718, opened 2022-05-09
Add functionality to set Algolia searchable and facet filters


RickyLB:
An alternative migrate function is to be created so the the api shouldn't have to rely on the current indexing version whenever there is new field added as a filter or searchable attribute.

PR #2739, opened 2022-05-14
feat(be/algolia): update algolia migrations

RickyLB:
closes https://github.com/ji-devs/ji-cloud/issues/2718

Algolia index settings will reset when the index_hash (derived from `hashfn` and stored in db) changes from the current hash of the index setting functions.

### DIFF
--- a/backend/api/src/algolia.rs
+++ b/backend/api/src/algolia.rs
@@ -12,7 +12,6 @@ use serde::Serialize;
 use std::collections::HashMap;
 use tracing::{instrument, Instrument};
 
-use hashfn;
 use shared::{
     domain::{
         asset::PrivacyLevel,
@@ -28,13 +27,15 @@ use sqlx::{types::Json, PgPool};
 use std::convert::TryInto;
 use uuid::Uuid;
 
-use migration::ResyncKind;
-
 mod migration;
 
 const PREMIUM_TAG: &'static str = "premium";
 const PUBLISHED_TAG: &'static str = "published"; // not currently used
 const HAS_AUTHOR_TAG: &'static str = "hasAuthor";
+
+pub const JIG_INDEX: &str = "jig_index";
+pub const MEDIA_INDEX: &str = "media_index";
+pub const COURSE_INDEX: &str = "course_index";
 
 #[derive(Serialize)]
 struct BatchJig<'a> {
@@ -84,7 +85,6 @@ struct BatchImage<'a> {
     tags: Vec<&'static str>,
     translated_description: &'a Vec<String>,
 }
-
 #[derive(Serialize)]
 struct BatchCourse<'a> {
     name: &'a str,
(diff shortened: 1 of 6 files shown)


backend/api/src/algolia.rs
```diff
@@ -116,6 +117,23 @@ enum BatchMedia<'a> {
...
+    CourseIndex,
+}
+
+impl AlgoliaIndices {
+    /// Represents the privacy level as a `str`. Relevant for Algolia tag filtering.
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::MediaIndex => "media_index",
+            Self::JigIndex => "jig_index",
+            Self::CourseIndex => "course_index",
```
johnnynotsolucky, line 132:
```suggestion
            Self::MediaIndex => MEDIA_INDEX,
            Self::JigIndex => JIG_INDEX,
            Self::CourseIndex => COURSE_INDEX,
```

backend/api/src/algolia/migration.rs
```diff
@@ -2,231 +2,135 @@ use algolia::{
...
-    media_index: &'a str,
-    jig_index: &'a str,
-    course_index: &'a str,
-) -> BoxFuture<'a, anyhow::Result<()>> {
-    // superceeded by `set_attributes_for_faceting_v5`
-    empty(client, media_index, jig_index, course_index)
-}
+pub(crate) const JIG_HASH: &'static str = JIG_INDEX_HASH;
+pub(crate) const MEDIA_HASH: &'static str = MEDIA_INDEX_HASH;
+pub(crate) const COURSE_HASH: &'static str = COURSE_INDEX_HASH;
```
johnnynotsolucky, line 16: No need for these lines since you're using the generated name https://github.com/ji-devs/ji-cloud/pull/2739/files#diff-a730c27e0999c83f61e3f67674ead1ffc1c16c829ab6a30b751c2752222a435fR46. 

If you're setting them so it's clearer that these exist, I added a [change](https://github.com/johnnynotsolucky/hashfn/commit/061a68bbc9631ef10b4d3aee7e233b592873dfc2) which lets you optionally set the name of the const using the macro:

```rust
#[hashfn(MEDIA_HASH)]
pub(crate) async fn media_index(...) {
  // ...
}
```

and then using it later on

```rust
sqlx::query!(
  r#"update algolia_index_settings set updated_at = now(), index_hash = $1 where index_name = $2"#, 
  MEDIA_HASH, 
  MEDIA_INDEX
)
  .execute(txn)
  .await?;
```

backend/api/src/algolia/migration.rs
```diff
@@ -2,231 +2,135 @@ use algolia::{
...
-    media_index: &'a str,
-    jig_index: &'a str,
-    course_index: &'a str,
-) -> BoxFuture<'a, anyhow::Result<()>> {
-    // superceeded by `set_attributes_for_faceting_v4`
-    empty(client, media_index, jig_index, course_index)
-}
+const JIG_INDEX: &str = "jig_index";
+const MEDIA_INDEX: &str = "media_index";
+const COURSE_INDEX: &str = "course_index";
```
johnnynotsolucky, line 12: Possible to reuse these from https://github.com/ji-devs/ji-cloud/pull/2739/files#diff-87eb355515c2b9aefffae7bab41c4c8616902b5fd2aae28effd8606c6af30dc7R35-R37 ?

johnnynotsolucky (approved):
Thanks @RickyLB, this looks good, it's great that the Algolia config is more readable!

I added a couple of comments, but nothing blocking.

PR #2760, opened 2022-05-18
fix(be/algolia): Remove multiple attributes for searchable attributes

johnnynotsolucky:
- Removes the multi attributes config for searchable attributes.

Since last week, the multiple attributes config has caused errors when searching against algolia:

image

The error is weird, but adding each of those attributes individually as searchable instead of comma separated fixes it.


johnnynotsolucky:
16542c963b2b492c70ddba577b590de02f43288d - Extra commit straight to sandbox because I missed the JIG index.

PR #2816, opened 2022-06-09
fix(be/image/search): correct order of image results from algolia

RickyLB:
This will return the same order of images that returns from Algolia instead of order by id.


corinnewo:
Is this based on relevance? How close it is to the search term?

Sent from my iPhone

Corinne Ossendryver
CPO, Jewish Interactive 
(+27) 079-8865326

Issue #2607, opened 2022-03-30
Content/Search: Age does not appear as tagged in curation tool.


chayaml:
The Jig below is tagged as suitable for pre-K - Kindergarten ages, but appears in the search as Pre-K - Pre-K, not sure why that is.
![Screen Shot 2022-03-30 at 2 07 06 PM](https://user-images.githubusercontent.com/78186896/160821152-a3f4d60d-2c89-4fc2-801e-b160cb8a6cfc.png)
![Screen Shot 2022-03-30 at 2 07 42 PM](https://user-images.githubusercontent.com/78186896/160821206-33e4879e-93a9-4699-bdb2-5e78f8b9451a.png)

corinnewo:
@mastapegs Please look into this. It should show 'Pre-K - K'. I'm also wondering whether the dash should not be replaced by a '>', as there are so many dashes when there is a Pre-K. 
@GaliShapira - what do you think of replacing the - with a >, as in 'Pre-K > K', or 'K > Elem'

GaliShapira:
Perhaps we can add from font awesome an arrow icon like one of these instead of the ">"? 
<img width="279" alt="Screen Shot 2022-03-30 at 16 07 39" src="https://user-images.githubusercontent.com/88834002/160841859-fa015316-1ba7-4f8e-a80d-2b2f1569991c.png">

corinnewo:
@GaliShapira -perfect!

corinnewo:
@johnnynotsolucky Please look into this. I assume it is a quick fix. I came across it again today with a JIG I created for the Ji Alef-Bet, and tagged it for Pre-K to K. We also asked that the dash showing the range would be an arrow instead.
![image](https://user-images.githubusercontent.com/16290361/168803085-5548a7db-11c7-45bb-a179-afe00bb51b32.png)

chayaml:
I just came across a Jig that I curated as suitable for K, but showed up in the search as suitable for Pre-K, so something is not right! @corinnewo 
![Screen Shot 2022-05-18 at 2 25 26 PM](https://user-images.githubusercontent.com/78186896/169118516-ecd58a77-6145-436b-8af6-a382a22728ea.png)
![Screen Shot 2022-05-18 at 2 30 40 PM](https://user-images.githubusercontent.com/78186896/169118701-2741ab64-8207-4d35-9bf4-987588039566.png)

PR #2818, opened 2022-06-11
fix(fe): Improved age range rendering; Copy updates

### LINKED ISSUE #2582: Delete a JIG on MY JIGs page - add a sentence to the warning
We have to make clear that deleting an item from MY JIGs deletes the item from the database. This warning is not enough. Need to add a sentence to warn that deleting from this list will delete altogether. (This differs from Ji Tap where you have a local creation area, and you can delete from there once you have published. So it is really important to add in the explanation.)

image

johnnynotsolucky:
Closes #2582, #2503 and #2607

- Adds a `content_html` property;
- Updates the delete JIG modal to include a new warning;
- Updates "Done" to "Generate" for card game lists;
- Updates the copy when a card games content hasn't been generated yet;
- Updates the age range so that it uses a FontAwesome arrow instead of `-`.

Screenshot 2022-06-10 at 12 36 45
Screenshot 2022-06-10 at 12 41 46
Screenshot 2022-06-11 at 20 27 03
Screenshot 2022-06-11 at 20 28 42

### DIFF
--- a/frontend/apps/crates/components/src/lists/dual/strings.rs
+++ b/frontend/apps/crates/components/src/lists/dual/strings.rs
@@ -1,4 +1,4 @@
-pub const STR_DONE: &str = "Done";
+pub const STR_DONE: &str = "Generate";
 pub const STR_CLEAR: &str = "Clear";
 
 pub mod error {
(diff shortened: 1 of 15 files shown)


MendyBerger:
@johnnynotsolucky can we discuss this later?

Issue #3190, opened 2022-09-14
Implement MIXPANEL


corinnewo:
Project token: bb8121928a7aac7a738924dce8c7f581

Recommend set up a proxy: https://developer.mixpanel.com/docs/collection-via-a-proxy
Javascript guide: https://developer.mixpanel.com/docs/javascript-quickstart?projectToken=bb8121928a7aac7a738924dce8c7f581

https://mixpanel.com/report/2798941/view/3333573/setup/?mp_source=intro-complete

## Events

- [x] Home page
  - [x] Buttons along the top menu – events to include the button name
    - [x] Home
	- [x] Content
	- [x] Create
	- [x] Community
	- [x] Classroom
	- [x] About Ji
	- [x] Donate
	- [x] Sign up
	- [x] Login
	- [x] Logout
- [x] Content
  - [x] Page load information:
	- [x] Search string
	- [x] Age
	- [x] Language
  - [x] Buttons on a search result card
	- [x] Play: include the name of the JIG/Resource/Course
	- [x] Share button
	- [x] Edit
- [x] Create
  - [x] Page load information
  - [x] ‘Create your own’ button
  - [x] On clicking a recent JIG at the bottom to edit it
  - [x] JIG builder page:
	- [x] When an activity is dragged into the JIG – need the name of the activity
	- [x] On Print cards button at the end of the card games
	- [x] Publish button in dialog
- [x] Player
  - [x] Page load information (if student or teacher player, name of JIG being played)
  - [x] If teacher panel:
	- [x] Button to open panel (click on head of the JIG)
	- [x] Share button
	- [x] Information button
- [x] Community
  - [x] Page load information
  - [x] Buttons – My profile, Circles, Members, ProDev

corinnewo:
@johnnynotsolucky Please can you try and implement this as soon as possible so that we can start tracking user behaviour - need this info for business plan and decisions regarding paywalls and subscriptions. Thanks!

PR #3192, opened 2022-09-16
feat(fe/mixpanel): Implement Mixpanel JS library; Add initial events

johnnynotsolucky:
Part of #3190


shared/rust/src/domain/asset.rs
```diff
@@ -124,6 +124,16 @@ pub enum AssetId {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::JigId(jig_id) => write!(f, "{}", &jig_id.0),
+            Self::CourseId(course_id) => write!(f, "{}", &course_id.0),
+            Self::ResourceId(resource_id) => write!(f, "{}", &resource_id.0),
+        }
+    }
+}
+
 impl From<JigId> for AssetId {
```
MendyBerger, line 137: I wouldn't impl display on something that's not meant to be displayed.

I'd rather use `.uuid().to_string()`, add a `.to_string()` method, or use `debug`.

Let me know if you disagree

frontend/apps/crates/utils/js/mixpanel.js
```diff
@@ -0,0 +1,6 @@
+import Mixpanel from 'mixpanel-browser';
+// Weird hacky solution. But I can't get wasm_bindgen to define a static_method_of Mixpanel without it
+// also attempting to create an import like `import {Mixpanel} from '...'`.
+//
+// So this just exports Mixpanel in a way that makes everyone happy.
+export {Mixpanel}
```
MendyBerger, line 6: Does this help you? https://rustwasm.github.io/docs/wasm-bindgen/print.html#js_name--blah

> The js_name attribute can also be used in situations where a JavaScript module uses export default. In this case, setting the js_name attribute to "default" on the type declaration, and the [js_class attribute](https://rustwasm.github.io/docs/wasm-bindgen/js_class.html) to "default" on any methods on the exported object will generate the correct imports.

frontend/apps/crates/components/src/page_header/dom.rs
```diff
@@ -55,6 +56,11 @@ pub fn render(
                 })
                 .property("href", &page_link.route())
                 .property("target", page_link.target())
+                .event(move |_evt: events::Click| {
+                    let mut properties = HashMap::new();
+                    properties.insert("Header Kind", page_link.kind_str().to_owned());
+                    mixpanel::track("Header Click", Some(properties));
+                })
```
MendyBerger, line 63: Are we going to have to call a function for each and every event to get mixpanel to work?

shared/rust/src/domain/asset.rs
johnnynotsolucky: I missed the `.uuid()` fn

frontend/apps/crates/utils/js/mixpanel.js
johnnynotsolucky: Nope. I've tried those. mixpanel's JS lib doesn't export a class like normal. It uses a singleton. So declaring the type and annotating with `js_name = default` and then the functions with `js_class = default` breaks the build. We need to use static_method_of so that we can call `mixpanel::<whatever>`, but that doesn't work with default. Hence this solution

frontend/apps/crates/components/src/page_header/dom.rs
johnnynotsolucky: This is similar to how we would do it in JS: `mixpanel.track('something', {prop: 'value'})`

MendyBerger: But I'm guessing it's going to be in each event handler?

PR #3300, opened 2022-10-27
feat(shared/macros): Add proc macro for testing the backend services

johnnynotsolucky:
NB: this will only work with a version of sqlx that includes the ::test macro (i.e. this branch I think: https://github.com/RickyLB/ji-cloud/tree/chore/update_actix)

Adds a new proc macro which wraps sqlx::test so that we can test our actix services and shut them down correctly without holding the db pools farmed out by sqlx::test.

- Test fns only have a single argument, `port: u16`, so that requests to the service can use the correct port.
- A service setup function needs to be provided with the following signature: `async fn setup_service(fixtures: &[T], services: &[T], pool_opts: PgPoolOptions, conn_opts: PgConnectOptions) -> (ServerHandle, u16)`
  - This setup function replaces the service setup logic that is used in every test.
- Uses catch_unwind to handle a panic so that the service can be shutdown correctly and release the db pool.

Note: Doesn't use sqlx::tests fixtures since we're already using a Fixture enum and its a ton of work to update that. But it would be easy enough to switch to using fixtures paths instead of enum variants- we'd only have to use `&[&'static str]` for the fixtures argument in the setup fn.

The Application struct should be updated to include a `server_handle` field, and _the `Drop` implementation needs to be removed_. I couldn't get it to call stop correctly/reliably.

```
pub struct Application {
    port: u16,
    server: Option<Server>,
    server_handle: actix_web::dev::ServerHandle, // <-- New
}

impl Application {
    pub fn new(port: u16, server: Server) -> Self {
        Self {
            port,
            server_handle: server.handle(),
            server: Some(server),
        }
    }

    pub fn port(&self) -> u16 {
        self.port
    }

    pub fn handle(&self) -> ServerHandle {
        self.server_handle.clone()
    }
    
    // ...
}
```

Example setup function:

```rust
async fn setup_service(
	fixtures: &[Fixture], 
	services: &[Service], 
	pool_opts: PgPoolOptions, 
	conn_opts: PgConnectOptions
) -> (ServerHandle, u16) {
    let app = initialize_server(fixtures, services, pool_opts, conn_opts).await;

    let handle = app.handle();
    let port = app.port();

    let _join_handle = tokio::spawn(app.run_until_stopped());

    (handle, port)
}
```

The macro takes three arguments:

- `setup` **Required**: The setup function, for example `setup = "my_service_setup"`;
- `fixtures` **Optional**: A list of `T` - Passed as the `fixtures` argument to the setup fn;
- `services` **Optional**: A list of `T` - Passed as the `services` argument to the setup fn.

```rust
#[test_service(
	setup = "setup_service", 
	fixtures("Fixture::User", "Fixture::Jig"), 
	services("Service::S3", "Service::Email"),
)
```

Example test case:

```rust
use macros::test_service;
use crate::{fixture::Fixture, service::Service};

//...

async fn setup_service(
    fixtures: &[Fixture], 
    services: &[Service], 
    pool_opts: PgPoolOptions, 
    conn_opts: PgConnectOptions
) -> (ServerHandle, u16) {
    // ... do setup

    (handle, port)
}

#[test_service(setup = "setup_service", fixtures("Fixture::User"))]
async fn create_default(
    port: u16, // <-- NB
) -> anyhow::Result<()> {
    let settings = insta::Settings::clone_current();

    let client = reqwest::Client::new();

    let resp = client
        .post(&format!("http://0.0.0.0:{}/v1/resource", port))
        .login()
        .send()
        .await?
        .error_for_status()?;

    assert_eq!(resp.status(), StatusCode::CREATED);

    let body: CreateResponse<ResourceId> = resp.json().await?;

    settings
        .bind_async(async {
            assert_json_snapshot!(body, {".id" => "[id]"});
        })
        .await;

    let resource_id = body.id.0;

    let resp = client
        .get(&format!(
            "http://0.0.0.0:{}/v1/resource/{}/draft",
            port, resource_id
        ))
        .login()
        .send()
        .await?
        .error_for_status()?;

    let body: serde_json::Value = resp.json().await?;

    insta::assert_json_snapshot!(
        body, {
            ".**.id" => "[id]",
            ".**.createdAt" => "[created_at]",
            ".**.lastEdited" => "[last_edited]"});

    Ok(())
}
```

### DIFF
--- a/backend/api/Cargo.lock
+++ b/backend/api/Cargo.lock
@@ -145,7 +145,7 @@ dependencies = [
  "actix-utils",
  "futures-core",
  "log",
- "mio",
+ "mio 0.7.13",
  "num_cpus",
  "slab",
  "tokio",
@@ -391,9 +391,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
+checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "backtrace"
@@ -1259,7 +1259,7 @@ checksum = "62007592ac46aa7c2b6416f7deb9a8a8f63a01e0f1d6e1787d5630170db2b63e"
 dependencies = [
  "futures-core",
  "lock_api",
- "parking_lot",
+ "parking_lot 0.11.2",
 ]
 
 [[package]]
@@ -1342,7 +1342,7 @@ dependencies = [
  "cfg-if",
  "js-sys",
  "libc",
- "wasi",
+ "wasi 0.10.0+wasi-snapshot-preview1",
  "wasm-bindgen",
 ]
 
@@ -1696,6 +1696,7 @@ dependencies = [
  "jsonwebtoken",
  "listenfd",
  "log",
+ "macros",
  "mp3-metadata",
  "once_cell",
  "openssl",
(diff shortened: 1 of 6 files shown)


MendyBerger:
@johnnynotsolucky can you please add code comments about what the motivation for this macro was, and how to use it?

johnnynotsolucky:
@MendyBerger done in #3314.

Issue #3307, opened 2022-10-28
Audio Instructions only (no written) - change the default text when timer, remove box if no timer


corinnewo:
When there is a **timer**, no written instructions, but there is audio, the default text says: "Are you ready?"
Please change to "1, 2, 3 Go!" so that non-English speaking children will understand it.

If there is **no timer**, and only audio without written instructions, no need for the box at all. Student can hear the instruction again by clicking the  replay instruction button in the player.

PR #3337, opened 2022-11-07
fix(fe/play): Prevent feedback advancing to next activity automatically when there is text and audio

johnnynotsolucky:
- Fixes issue when an activity has feedback configured with both text and audio, once the audio completed, the JIG would be advanced to the next activity.


johnnynotsolucky:
This is part of #3307.

Issue #3230, opened 2022-09-28
Flashcards - player - front and back of cards have different corner radius.


GaliShapira:
Can you make the front side of the card with more round corners so they look the same?
<img width="595" alt="Screen Shot 2022-09-28 at 11 31 35" src="https://user-images.githubusercontent.com/88834002/192806302-b3392a63-0182-49b7-971f-07851b5015e1.png">

PR #3385, opened 2022-11-22
fix(fe/flashcards): increase card border radius

RickyLB:
https://github.com/ji-devs/ji-cloud/issues/3230

## What?
Increases the border radius of the frontside of the flipped card

## Added
- selected element `.front.content`, and initialized `border-radius: 56px` for attributes `side="left"` and `size="flashcards`
- assigned variable `--border-radius: 16px` in `:host` to `border-radius` in `.content` element

### DIFF
--- a/frontend/elements/src/module/_groups/cards/play/card/card.ts
+++ b/frontend/elements/src/module/_groups/cards/play/card/card.ts
@@ -36,6 +36,10 @@ export class _ extends LitElement {
                     z-index: 1001;
                 }
 
+                :host([side="left"][size="flashcards"]) .front.content {
+                    border-radius: 56px;
+                }
+
                 section {
                     transition: transform 0.8s;
                     transform-style: preserve-3d;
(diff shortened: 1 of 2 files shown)


frontend/elements/src/module/_groups/cards/play/card/card.ts
```diff
@@ -36,6 +36,10 @@ export class _ extends LitElement {
                     z-index: 1001;
                 }
 
+                :host([side="left"][size="flashcards"]) .front.content {
```
johnnynotsolucky, line 39: @RickyLB this works, but not as expected. The cards in flashcards are reversed for some reason, so the left card has `side="right"` and the right card has `side="left"`. I should have noted this in the task. So this rule actually styles the wrong card correctly.

Instead of having this rule here, consider targeting just `"flashcards"` in styles.ts, and updating the new `--border-radius` css var that you added in that selector. That would make it so that all cards have a border radius of 16px, and just flashcards with the 56px radius.

frontend/elements/src/module/_groups/cards/play/card/card.ts
```diff
@@ -186,18 +190,18 @@ export class _ extends LitElement {
...
+                    styleKind,
+                    theme,
+                    mode,
+                    backSide
+                )}
                           >
                               <slot name="backSideContent"></slot>
                           </div>`
-                        : html`<img-ui
+                : html`<img-ui
```
johnnynotsolucky, line 204: nitpick: check formatting. 

We unfortunately don't have prettier set up yet to auto-correct these.

johnnynotsolucky (changes requested):
Nice work @RickyLB! 

I've left a couple of comments.

johnnynotsolucky (approved):
LGTM

PR #3392, opened 2022-11-28
fix(fe/instructions): increase gap between instruction and audio/upload texts

RickyLB:
https://github.com/ji-devs/ji-cloud/issues/3113

## Added 
- `div` has a styled grid
-  `row-gap` for `div` was increased to `16px`


johnnynotsolucky (approved):
Nice work @RickyLB! 

I made a mistake in my reading of the task. There should be less space between the radio buttons and the audio input, and more between the radio buttons and the field above: https://github.com/ji-devs/ji-cloud/issues/3113#issuecomment-1330201734

I think the top gap can be 20px, and the bottom around 4px.

You can merge once you've updated that 🤘

PR #3400, opened 2022-11-29
fix(fe/post_preview): add word activity to message

RickyLB:
resolves https://github.com/ji-devs/ji-cloud/issues/3394

## What 
1. All activity "done" pages have "activity" included in the "Your [activity] is ready!" message

## Added
1. In `post-preview.ts`, "activity" was prepended onto the existing `STR_HEADER_LINE_1_SUFFIX` value


johnnynotsolucky (approved):
LGTM

Issue #3397, opened 2022-11-29
Flashcards - make the image with the same corner radius of the frame


GaliShapira:
<img width="317" alt="Screen Shot 2022-11-29 at 17 23 20" src="https://user-images.githubusercontent.com/88834002/204570280-b7dec702-9e90-4c7b-b874-464c03f30ac4.png">

corinnewo:
@RickyLB I happened to choose an image with a white background, which now covers the new corner. Please can you apply the same radius to the image as well, so that this does not happen on the cards?

PR #3414, opened 2022-12-03
fix(fe/play/cards): resize oversized images in flashcards

RickyLB:
resolves https://github.com/ji-devs/ji-cloud/issues/3397

## Added
- `overflow` property, set to `hidden` to clip borders of oversized images


johnnynotsolucky (approved):
Nice 🤘

PR #3467, opened 2022-12-19
feat(fe/cover|poster): Adding 'next' play setting

### LINKED ISSUE #3284: Cover: Settings FINAL
Add in the settings tab, which is TOTALLY missing:

- [x] Add in option "Student views this page until..." and the options are 'clicking next' and 'after audio'
- [x] If there is audio recorded, then the default option is 'after audio'
- [x] If NO audio recorded, then force the 'clicking next' option.
- [x] Make sure that the 'clicking next' icon is an arrow like in the progress bar, and not the current continue button.
- [x] Add tooltip if there is no audio recorded, but click 'after audio': "This option available only when you have added audio."
- [x] Add in tab 'Instructions' to the cover settings, that will play before the audio of the content. (Most people won't have instructions on the cover, but there are use cases when teacher might want to give an instruction for the whole lesson...)
- [x] Add after-audio icon
- [x] Implement ending so that it moves to next activity based on new setting

@GaliShapira Please add in the Zeplin link here

### LINKED ISSUE #3285: Poster: Settings FINAL
In addition to #2629, here are the settings that should be in the poster:

- [x] Same continue as in cover, ie Add in option "Student views this page until..." and the options are 'clicking next' and 'after audio' (see ticket #3284)
- [x] Add in instructions tab, no feedback, as in cover.
- [x] Make sure that the instruction audio plays BEFORE the audio of the content.
- [x] Implement ending so that it moves to next activity based on new setting

@GaliShapira Please add in link to the zeplin.

johnnynotsolucky:
Closes #3284
Closes #3285

- Resolves issue where card games were not moving on to the next activity correctly.

### DIFF
--- a/frontend/apps/crates/components/src/module/_common/play/entry/ending/dom.rs
+++ b/frontend/apps/crates/components/src/module/_common/play/entry/ending/dom.rs
@@ -18,7 +18,7 @@ impl Ending {
 
         if let Some(kind) = state.kind {
             match kind {
-                ModuleEnding::Next => {
+                ModuleEnding::Next | ModuleEnding::Positive => {
                     log::info!("ending is next, transitioning...");
                     let msg = IframeAction::new(ModuleToJigPlayerMessage::Next);
 
--- a/frontend/apps/crates/components/src/module/_common/edit/settings/button/dom.rs
+++ b/frontend/apps/crates/components/src/module/_common/edit/settings/button/dom.rs
@@ -1,3 +1,5 @@
+use crate::overlay::handle::OverlayHandle;
+
 use super::state::*;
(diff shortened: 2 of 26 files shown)


frontend/apps/crates/components/src/module/_common/play/entry/ending/dom.rs
```diff
@@ -18,7 +18,7 @@ impl Ending {
 
         if let Some(kind) = state.kind {
             match kind {
-                ModuleEnding::Next => {
```
johnnynotsolucky, line 21: Resolves an issue when the ending is `Positive`, the activity does not move on to the next activity.

Issue #3399, opened 2022-11-29
Make Profile image editable again on My Settings page


corinnewo:
In 'My settings', can we add a pencil to the image to change the image?
https://zpl.io/vM54ynP

Currently, you can only access your profile in the Community section, under 'My Profile', but teachers do not think to find it there!

PR #3478, opened 2022-12-20
feat(fe/user/settings): edit profile image from My Settings 

RickyLB:
resolves https://github.com/ji-devs/ji-cloud/issues/3399

## Remaining
- [x] Use existing profile picture in popup window
- [x] Position pencil icon to top right corner

### DIFF
--- a/frontend/apps/crates/components/src/profile_image/actions.rs
+++ b/frontend/apps/crates/components/src/profile_image/actions.rs
@@ -13,7 +13,7 @@ use shared::{
 use utils::{prelude::ApiEndpointExt, unwrap::UnwrapJiExt};
 use web_sys::File;
 
-use super::{ProfileImage, ImageIdOrFile};
+use super::{ImageIdOrFile, ProfileImage};
 
 impl ProfileImage {
     pub fn apply_changes(self: &Rc<Self>) {
@@ -28,6 +28,8 @@ impl ProfileImage {
             };
 
             (state.config.save_changes)(image_id);
+
+            state.popup_open.set(false)
         }));
     }
 }
--- a/frontend/apps/crates/components/src/profile_image/actions.rs
+++ b/frontend/apps/crates/components/src/profile_image/actions.rs
@@ -1,0 +1,50 @@
+use std::rc::Rc;
+
+use crate::image::upload::upload_image;
+use dominator::clone;
+use shared::{
+    api::endpoints,
+    domain::image::{
+        user::{UserImageCreatePath, UserImageCreateRequest},
+        ImageId, ImageSize,
+    },
+    media::MediaLibrary,
+};
+use utils::{prelude::ApiEndpointExt, unwrap::UnwrapJiExt};
+use web_sys::File;
+
+use super::{ProfileImage, ImageIdOrFile};
+
+impl ProfileImage {
+    pub fn apply_changes(self: &Rc<Self>) {
+        let state = self;
+        state.loader.load(clone!(state => async move {
+            let image_id = match state.image.get_cloned() {
+                None => None,
+                Some(ImageIdOrFile::ImageId(image_id)) => Some(image_id),
+                Some(ImageIdOrFile::File(image_file)) => {
+                    Some(upload_profile_image(image_file).await.unwrap_ji())
+                },
+            };
+
+            (state.config.save_changes)(image_id);
+        }));
+    }
+}
+
+async fn upload_profile_image(file: File) -> Result<ImageId, Box<dyn std::error::Error>> {
+    let req = UserImageCreateRequest {
+        size: ImageSize::UserProfile,
+    };
+
+    let image_id = endpoints::image::user::Create::api_with_auth(UserImageCreatePath(), Some(req))
+        .await
+        .map_err(|_err| "Error creating image in db")?
+        .id;
+
+    upload_image(image_id, MediaLibrary::User, &file, None)
+        .await
+        .map_err(|_err| "Error uploading image")?;
+
+    Ok(image_id)
+}
--- a/frontend/apps/crates/components/src/profile_image/config.rs
+++ b/frontend/apps/crates/components/src/profile_image/config.rs
@@ -1,0 +1,20 @@
+use shared::domain::{image::ImageId};
+
+pub const STR_WRONG_FILE_TYPE: &str = "Oh no! We don't accept this type of file.";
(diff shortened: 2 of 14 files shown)


frontend/apps/crates/components/src/profile_image/styles.css
```diff
@@ -0,0 +1,73 @@
+.edit-button {
+    grid-row: 1;
+    grid-column: 1;
+    position: relative;
+    left: 60px;
```
MendyBerger, line 5: Instead of using positioning with a hardcoded 60 pixels, you should probably use the following:
```css
justify-self: end;
align-self: start;
```

Also, make sure that the button is at least 32 pixels in size. (needed for touch, fingers are larger than cursors)

```css
min-width: 30px;
min-height: 30px;
```

frontend/apps/crates/entry/user/src/settings/callbacks.rs
```diff
@@ -0,0 +1,6 @@
+use shared::domain::user::UserProfile;
+
+pub struct EditProfileCallbacks {
+    pub save_changes: Box<dyn Fn(UserProfile)>,
+    pub close: Box<dyn Fn()>,
+}
```
MendyBerger, line 6: Don't think this is needed, already have `ProfileImageConfig`

frontend/apps/crates/entry/user/src/settings/mod.rs
```diff
@@ -1,3 +1,4 @@
 pub mod actions;
 pub mod dom;
 pub mod state;
+pub mod callbacks;
```
MendyBerger, line 4: Can get rid of this I think, like `EditProfileCallbacks`

frontend/apps/crates/components/src/profile_image/config.rs
```diff
@@ -0,0 +1,20 @@
+use shared::domain::{image::ImageId};
+
+pub const STR_WRONG_FILE_TYPE: &str = "Oh no! We don't accept this type of file.";
+pub const STR_FILE_TOO_LARGE: &str = "Oh no! This file is too heavy. Maximum file size: ";
+
+pub struct ProfileImageConfig {
+    pub close: Box<dyn Fn()>,
```
MendyBerger, line 7: Do we need the close callback? Don't think it's ever called

frontend/apps/crates/entry/user/src/settings/dom/mod.rs
```diff
@@ -44,20 +44,22 @@ impl SettingsPage {
...
+            .child(ProfileImage::new(state.user.profile_image.read_only(), 
+                ProfileImageConfig {
+                    close: Box::new(clone!(state => move || {
+                        state.active_popup.set(ActivePopup::None);
+                        log::info!("In Close")
+                    })), 
+                    save_changes: Box::new(clone!(state => move |user| {
+                        state.user.profile_image.set(user);
+                        state.save_profile();
+                        log::info!("Save Change")
```
MendyBerger, line 56: Logging is only needed for debugging, you can get rid of it

frontend/apps/crates/components/src/profile_image/config.rs
```diff
@@ -0,0 +1,20 @@
+use shared::domain::{image::ImageId};
+
+pub const STR_WRONG_FILE_TYPE: &str = "Oh no! We don't accept this type of file.";
+pub const STR_FILE_TOO_LARGE: &str = "Oh no! This file is too heavy. Maximum file size: ";
```
MendyBerger, line 4: Don't think we need these constants

frontend/apps/crates/components/src/profile_image/dom.rs
```diff
@@ -0,0 +1,152 @@
...
+                            clone!(state => move || {
+                                render_popup(&state)}),                        
+                                Some(Box::new(clone!(state => move || {
+                                    state.popup_open.set(false);
+                                })))
+                        ))
+                    } else {
+                        Some(Box::new(clone!(state => move || {
+                            state.popup_open.set(false);
+                        })));
```
MendyBerger, line 58: Don't think this `Some` does anything, it's never used

shared/rust/src/domain/user.rs
```diff
@@ -293,6 +293,42 @@ pub struct UserProfileExport {
     pub opt_into_edu_resources: bool,
 }
 
+impl Default for UserProfile {
```
MendyBerger, line 296: Please get rid of this as we discussed

shared/rust/src/domain/user.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
 pub mod public_user;
 
 /// Wrapper type around [`Uuid`], represents the ID of a User.
-#[derive(Copy, Clone, Eq, PartialEq, Serialize, Deserialize, Debug, PathPart)]
+#[derive(Copy, Clone, Eq, PartialEq, Serialize, Deserialize, Debug, PathPart, Default)]
```
MendyBerger, line 21: Please get rid of `Default` as we discussed

frontend/apps/crates/components/src/profile_image/actions.rs
```diff
@@ -0,0 +1,50 @@
...
+        state.loader.load(clone!(state => async move {
+            let image_id = match state.image.get_cloned() {
+                None => None,
+                Some(ImageIdOrFile::ImageId(image_id)) => Some(image_id),
+                Some(ImageIdOrFile::File(image_file)) => {
+                    Some(upload_profile_image(image_file).await.unwrap_ji())
+                },
+            };
+
+            (state.config.save_changes)(image_id);
```
MendyBerger, line 30: Think the popup should be closed after saving, something like `state.popup_open.set(false)`

MendyBerger (changes requested):
Great work @RickyLB!

Adding a few comments, but great looks great overall!

Make sure to format the code with `cargo fmt`, and to check for warnings with `cargo check`.

Please also use this new component in the community my profile page, instead of having the same implementation twice.

frontend/apps/crates/components/src/profile_image/config.rs
RickyLB: Agreed!

MendyBerger:
Great job @RickyLB!
I'm going to merge this, but please also use this component in the my-profile page in community

MendyBerger:
@RickyLB  #3500

PR #3481, opened 2022-12-21
feat(fe/tapping-board): Only highlight tapped traces while audio is playing

### LINKED ISSUE #3309: Listen and learn - don't keep clicked items highlighted - highlight ONLY while audio is playing
Currently, all clicked areas in the listen and learn remain highlighted after they have been clicked. This is confusing for many reasons, the main one being that when the audio plays, the child does not know which highlight is playing. Also, the screen can get really crowded!

Please change so that when you click an item to hear it, the highlight appears for the duration of the audio, and then disappears.

johnnynotsolucky:
Closes #3309

### DIFF
--- a/frontend/apps/crates/entry/module/tapping-board/play/src/base/game/playing/actions.rs
+++ b/frontend/apps/crates/entry/module/tapping-board/play/src/base/game/playing/actions.rs
@@ -2,19 +2,27 @@ use super::state::*;
 use components::module::_common::play::prelude::*;
 use shared::domain::module::body::tapping_board::Next;
 use std::rc::Rc;
+use utils::unwrap::UnwrapJiExt;
 
 impl PlayState {
     pub fn select(state: Rc<Self>, index: usize) {
+        // Kill playback for all already selected traces
+        state
+            .selected_set
+            .lock_ref()
+            .iter()
+            .for_each(|index| state.traces.get(*index).unwrap_ji().kill_playback());
+
         // mark the selected set
         state.selected_set.lock_mut().insert(index);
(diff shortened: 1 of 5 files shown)


frontend/apps/crates/entry/module/tapping-board/play/src/base/game/playing/actions.rs
```diff
@@ -2,19 +2,27 @@ use super::state::*;
...
 
 impl PlayState {
     pub fn select(state: Rc<Self>, index: usize) {
+        // Kill playback for all already selected traces
+        state
+            .selected_set
+            .lock_ref()
+            .iter()
+            .for_each(|index| state.traces.get(*index).unwrap_ji().kill_playback());
+
```
johnnynotsolucky, line 15: Small optimization: Instead of iterating over every trace, only end playback of traces which are in the `current_set`

frontend/apps/crates/entry/module/tapping-board/play/src/base/game/playing/actions.rs
```diff
@@ -2,19 +2,27 @@ use super::state::*;
...
-            } else {
-                trace.kill_playback();
-            }
+        {
+            let mut current_set = state.current_set.lock_mut();
+            current_set.clear();
+            current_set.insert(index);
         }
+
+        state.traces.get(index).unwrap_ji().select(state.clone());
```
johnnynotsolucky, line 25: Same as above, no need to iterate over entire trace Vec - we can get the trace and call select directly.

Issue #3480, opened 2022-12-21
User table for admin


corinnewo:
Need to add a table to the admin panel (https://jigzi.org/admin) of all the users that have signed up.
You can see Gali's original plan for this table here: https://zpl.io/XYn76QP - we are simplifying it. We will be adding functionality later, but initially we need a table of users that we can search.

**Columns needed:**

- Username
- First name
- Last name
- Email (which will be clickable to open an email client)
- Country
- State
- City
- School/Organization
- No. of Creations (a count of how many JIGs, resources, courses that they have created) (sortable) - this will help us see the most active users easily
- Date signed up (sortable)
- Language (what they chose for newsletters - I think it is 'language of communication')

**Search to include these fields:**

- first name
- last name
- School/org
- country
- state
- date joined 

Future functionality:
- export current search results to csv
- filter according to a date range
- **Delete** a user
- **Reset** password for a user
- **Edit** credentials (especially school)
- Order/filter columns (if not done in the initial table)
- Click the name, and can see their community profile (in a new tab/popup?)

corinnewo:
@RickyLB Please talk to @MendyBerger before you start with this! Thanks

corinnewo:
@RickyLB Please add in a count of the number of results found. For example, if I search for 'Chicago', then display along the top "28 users found". Thanks!

PR #3519, opened 2023-01-04
feat(fe/admin/users): add basic user table to admin

RickyLB:
https://github.com/ji-devs/ji-cloud/issues/3480

## Added
- `Users` tab
- table/list of basic user info (names)
## Fields
- `Username`
- `First Name`
- `Last Name`
## Note
- This is a temporary implementation for `Users` in the admin table
## Screenshot
admin user table


RickyLB:
@MendyBerger Originally, I had gotten rid of the `curation` keyword in `admin-curation-table` element but using `admin-table-*` for both curation and users yielded conflicting elements. I added `user` and `curation` back into the element names.

frontend/apps/crates/entry/admin/src/users/actions.rs
```diff
@@ -0,0 +1,129 @@
...
+    routes::{AdminRoute, AdminUsersRoute, Route},
+};
+
+use super::{editable_user::EditableUser, FetchMode, Users};
+
+impl Users {
+    pub fn load_data(self: &Rc<Self>) {
+        let state = Rc::clone(self);
+        state.loader.load(clone!(state => async move {
+            join!(
```
MendyBerger, line 26: No need to `join!` if you only have one future

frontend/apps/crates/entry/admin/src/users/user/state/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+}
+
+impl AdminUser {
+    pub fn new(users_state: Rc<Users>, user_id: UserId, user: Rc<EditableUser>) -> Rc<Self> {
+        Rc::new(Self {
+            user_id,
+            user,
+            loader: AsyncLoader::new(),
+            users_state,
+            player_open: Mutable::new(false),
```
MendyBerger, line 23: There's no player here, you can get rid of this

frontend/apps/crates/utils/src/routes.rs
```diff
@@ -660,6 +673,13 @@ impl From<&Route> for String {
                         format!("/admin/curation/{}", jig_id.0)
                     }
                 },
+                AdminRoute::Users(users_route) => match users_route {
+                    AdminUsersRoute::Table => "/admin/users".to_string(),
+                    AdminUsersRoute::User(user_id) => {
+                        log::info!("routes.rs user: {:?}", user_id);
```
MendyBerger, line 679: I'm guessing this is left over from debugging

MendyBerger (commented):
@RickyLB this looks great! 

About the element name conflict, I meant that you should use the same element for both tables, not define another element with the same name.

PR #3743, opened 2023-03-10
chore(be): Use crates.io sparse protocol for building images

johnnynotsolucky:
Could shave up to a minute off the build time.

~Will merge once the v1.68.0 base docker image is available.~


johnnynotsolucky:
Shaved ~67s.

Super scientific measurements:

Before:
image

After:
image

PR #4159, opened 2025-08-12
Tweaks  small footer updates

duncan1a:
1. Quick tour - please make the video open in a new tab, and not overwrite Jigzi.
2. Ji Tutorials - Please change to "Jigzi tutorials' and link to https://www.jewishinteractive.org/how-to-jigzi/ in a new tab


duncan1a:
Hi @johnnynotsolucky  - those are just lock-file updates, probably from when I initially did `npm i` with node 15 and then later again with node 20. It doesn't actually change the js . I can remove them, but it is generally recommended to include them

johnnynotsolucky:
Reverting this. The package lock changes removed required dependencies.

PR #4162, opened 2025-08-20
Remove tips and featured jigs for non logged in users

duncan1a:
Hide all the extra rows  (Trending and Top Picks)so that a new user gets to see the animation about jigzi first.

I also removed a slight space on the `empty-fragment` element of the wix segment above it.

@johnnynotsolucky @corinnewo


duncan1a:
@johnnynotsolucky there's one check failing. Should I build something before committing? I tried `cargo build` in the frontend apps dir.This update works fine locally

PR #4165, opened 2025-09-04
Make empty search results hint conditional

duncan1a:
If a user searches and there are no results, a message comes up saying - can't find matches, try selecting 'all' 
But this needs to be changed now, as we changed the default to "All"

Solution: Only have the suggestion showing if the tab is set to 'Top rated'. If the tab is set to 'All', the message should be only the first sentence: "Oh snap! We couldn't find any matches for [string]."

Rated
image

All
image

With results works as normal
image

@johnnynotsolucky @corinnewo


duncan1a:
@johnnynotsolucky could you help me out here? I had a similar diff error before with a PR and couldn't figure it out.

This one (https://github.com/ji-devs/ji-cloud/actions/runs/17474069727/job/49629437759?pr=4165) has two diffs. The first one doesn't seem to have any difference. The second one is only formatting. Both of them have no practical difference in code - it's just formatting - so should have no impact on working code at all. Also, both of these are new additions, so I can't figure out what the diff is against.


image

